PR #3, opened 2019-08-29
Add ticket model, synchronizer

kti-cameron:


djautotask/sync.py
```diff
@@ -0,0 +1,219 @@
...
+        self.synced_ids = set()
+
+
+class Synchronizer:
+    lookup_key = 'id'
+
+    def __init__(self, full=False, *args, **kwargs):
+        self.full = full
+
+        self.at_api_object = self.init_api_connection()
```
kti-sam: Something to consider, don't know if its worth caring about maybe Matt can comment:

> but if you are initialising the library often, it may pay to also include the zone url, otherwise it needs to be discovered by performing an API lookup

atws seems to indicate that connecting multiple times will increase the overhead of syncing, so maybe we want to make the connection once for a 'sync session', or provide a zone like they suggest.

kti-sam: Also I think naming it just at_api might be good. I feel the 'object' at the end is redundant here. Mabye client instead of object?

djautotask/tests/fixture_utils.py
```diff
@@ -0,0 +1,31 @@
...
+    ticket_object.id = 7703
+    ticket_object.Title = 'Customer defect with provisioning'
+    ticket_object.CompletedDate = '2019-08-29 19:01:41.573000+01:00'
+    ticket_object.CreateDate = '2012-06-18 13:38:30.940000+01:00'
+    ticket_object.Description = \
+        'Review with customer how they can utilize new features.'
+    ticket_object.DueDateTime = '2012-06-18 13:38:30.940000+01:00'
+    ticket_object.EstimatedHours = 3.0
+    ticket_object.LastActivityDate = '2019-08-29 00:16:31.970000+01:00'
+    ticket_object.TicketNumber = 'T20120529.0001'
```
kti-sam: As we discussed, since they named these attributes like this we can actually use the fixture keys to `setattr` to make this very generic.

kti-sam:
Since I messed up the integration code I actually discovered the errors are formatted a bit wonky:

`Failed to sync Ticket: 'Autotask API has returned an error.\n    Please process the response attribute in this exception to determine\n    the extent of this failure.\n    '[IntegrationCode is invalid., '<queryxml><entity>Ticket</entity><query><field>LastActivityDate<expression op="GreaterThanorEquals">2019-09-05 16:18:02</expression></field></query></queryxml>']`

See how its asking you to process the response, and the newlines distributed throughout.

kti-sam:
And of course the commands tests are missing, but you mentioned that I see now.

djautotask/sync.py
kti-cameron: Yeah object is not necessary. I changed it to client. 

I'm not really sure about the best for the sync overhead point. I think its probably something that might be better to discuss later. But I've added env settings to set the api version and url to use (providing a url skips the zone api request).

djautotask/tests/fixtures.py
```diff
@@ -0,0 +1,41 @@
+
+API_SERVICE_TICKET = {
+    'AccountID': 29683561,
+    'AllocationCodeID': 29683407,
+    'AssignedResourceID': 29683794,
+    'AssignedResourceRoleID': 29683396,
+    'ChangeInfoField1': '',
+    'ChangeInfoField2': '',
+    'ChangeInfoField3': '',
+    'ChangeInfoField4': '',
```
kti-sam: Is it common for several of these fields to come in blank in real requests? If so just leave as is, if not, remove fields we wouldn't normally receive as empty strings.

djautotask/tests/fixture_utils.py
```diff
@@ -0,0 +1,42 @@
...
+        setattr(suds_object, key, value)
+
+    return suds_object
+
+
+def generate_objects(object_type, fixture_objects):
+    """
+    Generate multiple objects based on the given fixtures.
+    """
+    path = os.path.abspath("djautotask/tests/atws.wsdl")
```
kti-sam: I think just a quick explanaition in a comment of what this atws.wsdl is, and why we are using it would be good here for posterity.

djautotask/tests/test_commands.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_sync(self):
+
+        out = self._test_sync(*self.args)
+        obj_title = self._title_for_at_object(self.args[1])
+
+        self.assertIn(obj_title, out.getvalue().strip())
+
+    def test_full_sync(self):
+        self.test_sync()
```
kti-sam: This looks similar to something you fixed in django-connectwise recently where we were running tests multiple times when we didn't intend to, causing errors. Could this cause the same problems or is this different. I see we didn't fix this in the abstractbasesynctest in djc eitehr though.

kti-sam:
Ok, that's my review for these changes.

djautotask/tests/fixtures.py
kti-cameron: Based on the ticket from our sandbox yes, I've only seen them empty. The docs are not clear on what these fields are actually used for. We are not using them now so I can just remove them.

kti-sam: No no, do they show up in the replies as empty strings, or do they not show up at all? If they show up as empty strings from the data we receive leave them in the mock data. We want to keep as close to reality as possible.

kti-cameron: Yes they are returned as empty stings. I'll leave them in the fixture.

kti-cameron: As for the unclosed file warning. I'm not seeing the warning myself. I think its happening in fixture_utils.py: line 35 when the client object is instantiated. I found an open issue for it at https://bitbucket.org/jurko/suds/issues/102/resourcewarning-unclosed-file-with-python

I'm not sure how to resolve this since the suds library is actually opening the file not us. Otherwise I could just add a proper `with ... open():` etc. Maybe there's something else I'm missing here.

djautotask/tests/test_commands.py
kti-cameron: Yeah this is a little similar but the other case was mostly just a lot of poor naming which I thought was resulting in tests running when they shouldn't. I've improved the naming a bit but I think this should be ok for now. 

Might need to change it a bit more when we start adding more classes to sync.

djautotask/tests/fixtures.py
kti-sam: Yeah I'm consistently getting this, even on my home workstation. I'll show you tomorrow morning. If you can fix it that simply great. I think this is the only thing left before merging.

kti-sam:
Yeah, just the open file warning left. I'll show you what I am seeing tomorrow morning and we can try and figure out why you aren't seeing it, (or why only I am haha)

djautotask/tests/fixtures.py
kti-cameron: Because the suds Client is opening the file it limited my options for fixing this unless we open a PR for suds-jurko. I dug around in the suds library and found a possible location for the file not being closed but not sure because we only get the warning once not every time we open the file. So not really sure if I know exactly the location that this. 

Instead of fixing the issue directly, I've moved the client instantiation code to its own function and set as a constant so that the file only gets opened once. This removes the warning but I'm guessing the issue is likely still there. At the very least its probably better to open the file once instead of repeatedly.

kti-sam: I think that should be fine, I am not seeing the warning either, and everything is still working.

PR #5, opened 2019-09-16
1062 add resource model and sync to django autotask

kti-cameron:


djautotask/sync.py
```diff
@@ -320,11 +311,26 @@ def _assign_field_data(self, instance, object_data):
         self.set_relations(instance, object_data)
         return instance
 
-    def get_related_instance(self, relation_id):
-        return models.TicketStatus.objects.get(value=relation_id)
+    def get_related_instance(self, relation_id, object_field):
```
kti-sam: I might be blind but I don't see anything calling this method

kti-cameron: Nope, that should have been removed. Fixed now.

PR #6, opened 2019-09-16
1063 add ticketsecondaryresource model and sync to django autotask

kti-cameron:


kti-sam:
Looks pretty good, merge waiting on response to review in 1062 and concern mentioned in gitlab issue for 1062

PR #7, opened 2019-09-17
1064 add priority and queue model and sync to django autotask

kti-cameron:


kti-sam:
I didn't really realize that the picklist ID's were the same as the value. I think if we are syncing picklists so the primary key is the same as the value we should just drop tracking the value as a field completely. As long as I am correct in assuming that the 'value' on our picklists is always going to be an integer, right?

kti-cameron:
Resolved conflicts. Ready for another look

djautotask/tests/test_commands.py
```diff
@@ -85,32 +85,59 @@ def test_full_sync(self):
...
-        fixtures.API_TICKET_STATUS_LIST,
-        'ticket_status',
+        mocks.ticket_priority_api_call,
+        fixtures.API_TICKET_PRIORITY_LIST,
+        command_name,
+    )
+
+
+class TestSyncQueueCommand(AbstractPicklistSyncCommandTest, TestCase):
+    command_name = 'queue'
```
kti-sam: There doesn't seem to be a good reason to assign the command name here. I would expect this to be called somewhere like `cls` or `self`, but I can't find any usages anywhere besides in the `args` assignment. 

I also don't see this pattern in django connectwise. So take this out from all the models, unless you do have a good reason for this and I am just not seeing it.

kti-sam:
Just that one nitpick and it should be good

djautotask/tests/test_commands.py
kti-cameron: Hmm, I think that was a leftover that should've been removed. I can't recall the reason for it but I resolved another way and never removed the `command_name` field. Thanks for spotting it. Fixed now

PR #8, opened 2019-09-17
1065 add account company model and sync to django autotask

kti-cameron:


kti-sam:
Looks good I think, waiting on conflicts

djautotask/tests/mocks.py
```diff
@@ -34,6 +34,10 @@ def secondary_resource_api_call(return_value):
     return create_mock_call(WRAPPER_QUERY_METHOD, return_value)
 
 
+def account_api_call(return_value):
+    return create_mock_call(WRAPPER_QUERY_METHOD, return_value)
```
kti-sam: These are all the same now that this is growing. I guess make parent classes for picklist/non-picklist

kti-cameron: Yes, good point. I'm actually thinking that have a function for each type of API call is unnecessary here anyway since they're all mocking the method (`atws.wrapper.Wrapper.query` or `atws.helpers.get_field_info`) unlike with the CW API where they have different endpoints. 

What if we just have two functions, one for mocking picklist calls and one for the normal query calls?

kti-sam: Oh yes, its not even adding anything per model to separate it. Thats a great idea, yeah, do that.

kti-cameron: Changes made. Ready for another look.

PR #9, opened 2019-09-18
1066 add project model and sync to django autotask

kti-cameron:


djautotask/models.py
```diff
@@ -52,13 +71,103 @@ class Picklist(TimeStampedModel):
...
+class Project(TimeStampedModel):
+    name = models.CharField(max_length=100)
+    number = models.CharField(null=True, max_length=50)
+    description = models.CharField(max_length=2000)
+    actual_hours = models.DecimalField(
+        null=True, decimal_places=2, max_digits=9)
+    completed_date_time = models.DateTimeField(null=True)
+    completed_percentage = models.PositiveSmallIntegerField(default=0)
+    duration = models.PositiveSmallIntegerField(default=0)
+    start_date_time = models.DateTimeField(null=True)
```
kti-sam: Huh. They really want to track their projects closely I guess. "THE PROJECT WILL START FRIDAY AT EXACTLY 2:54 PM! AND NOT A MINUTE SOONER!"

kti-sam: Actually no, its not. I checked autotask and it looks like they display it as a date only and ignore the time even though it comes in as a date time. We can probably store it as a date field too then just like we do in djconnectwise. Unless there is a good reason not to. I feel like ambiguous project start/end dates based on time zones is not good though.

djautotask/models.py
```diff
@@ -52,13 +71,103 @@ class Picklist(TimeStampedModel):
...
+
+
+class ProjectStatus(Picklist):
+    pass
+
+    class Meta:
+        verbose_name_plural = 'Project statuses'
+
+
+class ProjectType(Picklist):
```
kti-sam: Now this and status were not technically part of this issue, but I can't think of a good enough reason to say take it out at this point... so :woman_shrugging:

kti-cameron: Yeah I know. I'm guilty. I figured they'll be useful later on and they were super easy to add while I was doing everything else.

djautotask/models.py
kti-cameron: Good point. Might as well take the same approach as django-connectwise. :+1:

kti-cameron:
Alright, resolved the conflicts. Ready for another look

kti-cameron:
Actually not ready yet. Still need to fix some sync date parse issues.

PR #10, opened 2019-09-24
1067 add a number of picklists models and fks to django autotask

kti-cameron:
Added several picklist fields.

Couple things to note:
*  I realized that I broke the tests in the last issue with some changes to how the dates in the fixtures are parsed. I fixed it in this issue by storing the dates as datetimes in the fixtures. 
* I also fixed another issue when syncing picklist entities. They would update and then get deleted on full syncs even though they hadn't been removed from Autotask. 

In sync.py: 182 
```python 
stale_ids = initial_ids - synced_ids
```
synced_ids was not of type int and so the when difference was calculated it would return the full set of picklist records and remove them all.


kti-sam:
Took a little longer to review because I wanted to be more thorough with the review. Checking syncs multiple times and going through each model, etc. Looks good

PR #11, opened 2019-09-30
1068 get authorization and network timeouts config from main project same as django connectwise

kti-cameron:


djautotask/sync.py
```diff
@@ -62,22 +68,89 @@ def __init__(self):
         self.synced_ids = set()
 
 
+class CustomRequestsTransport(transport.Transport):
```
kti-sam: I like how you did this. So I went through the synchronizer's `init_api_connection` method and I see the only time self is used is to reference `request_settings`. I think moving CustomRequestsTransport to its own file and much, or all, of the `init_api_connection` method on to this class is a good idea. That way we can keep more of a sync/api separation like we do in django-connectwise.

djautotask/api.py
```diff
@@ -0,0 +1,89 @@
...
+from io import BytesIO
+import suds.transport as transport
+from atws.wrapper import AutotaskAPIException, ResponseQuery
+from atws import wrapper, connection
+
+from django.conf import settings
+from djautotask.utils import DjautotaskSettings
+
+
+def init_api_connection(**kwargs):
```
kti-cameron: I didn't include this as a class method on AutotaskRequestsTransport. Didn't make sense to me because `init_api_connection` returns a `Wrapper` class with a reference to `AutotaskRequestsTransport`.

kti-sam: Oh yeah, that makes sense. Looks good to me.

PR #13, opened 2019-10-17
let awts get url automatically

kti-sam:


kti-sam:
This is now obsolete, we will be providing the URL manually to set up so django-kanban can access the region URL as well.

PR #16, opened 2019-10-23
Improve wsdl path finding

kti-cameron:
Added a small improvement to building the path for the wsdl. This way we can use it from the main app for easily generating test AT objects.


kti-sam:
This needs a bump to .13 now

kti-sam:
Make sure to pull in master or you will get a merge conflict

PR #18, opened 2019-10-29
add callback support for autotask

kti-sam:


kti-sam:
That was annoying... Quite the setback after 'the incident'. One day I'll remember to add unit tests.

PR #19, opened 2019-10-30
1178 sync license type to resource model

kti-cameron:
Added license type model and synchronizer. I also noticed that the ticket search in django admin was broken so fixed that too.


kti-sam:
Getting an exception when trying to do a full sync:
```
  File "/home/sam/git/Kanban/env3/lib/python3.5/site-packages/django_autotask-0.0.13a0-py3.5.egg/djautotask/sync.py", line 256, in get_batch_queries
    min_id = max_id
UnboundLocalError: local variable 'max_id' referenced before assignment
```

Looks like in `sync.py` there can be a case where `max_id` doesn't get initialized before the end of the loop.

kti-sam:
Oh haha, I didn't see this comment. Looks like we are both seeing that, which means its in master.

kti-cameron:
Bumped version number. Took two commits to do it though :man_facepalming:

PR #20, opened 2019-10-31
1180 unboundlocalerror on syncs in django autotask

kti-cameron:
Added the migrations merge.

I forgot that I'd committed a fix for the reference error so I'll keep it here as it should allow us to sync tickets again. 

I'll add a new issue to fix syncing more than 500 records.


kti-cameron:
Updated again to only include the migration merge as discussed.

kti-cameron:
No longer required. Can be closed now.

PR #30, opened 2019-11-25
1173 sync autotask project tasks

kti-cameron:


djautotask/admin.py
```diff
@@ -132,3 +132,14 @@ class AccountAdmin(admin.ModelAdmin):
 class ProjectAdmin(admin.ModelAdmin):
     list_display = ('id', 'name', 'number')
     search_fields = ('id', 'name', 'number')
+
+
+@admin.register(models.Task)
+class TaskAdmin(admin.ModelAdmin):
+    list_display = ('id', 'title', 'number')
```
kti-sam: For Connectwise I found it useful to have the status available for a ticket-like object to be in the list view, so it is available at a glance, or filtering. Please add that here. Actually on second thought maybe as a search field too.

djautotask/models.py
```diff
@@ -271,3 +271,44 @@ class Project(TimeStampedModel):
 
     def __str__(self):
         return self.name
+
+
+class Task(TimeStampedModel):
+    title = models.CharField(blank=True, null=True, max_length=255)
```
kti-sam: Good news, I double checked the limits on title and description and it appears they are actually telling the truth this time.

djautotask/sync.py
```diff
@@ -282,14 +282,31 @@ def _assign_field_data(self, instance, object_data):
...
+    def _get_query_conditions(self, query):
+        # Don't sync the 'Complete' status for tickets and tasks.
+        # Most if not all tickets/tasks end up here and stay here forever
+        # or until deleted.
+        # This would cause ticket/task syncs to take a very long time.
+        query.open_bracket('AND')
+        query.WHERE(
+            'Status',
+            query.NotEqual,
+            self.at_api_client.picklist['Ticket']['Status']['Complete']
```
kti-sam: I can't remember if you told me already, did we confirm this is actually ok to do? Even though it is technically ticket? In theory it makes sense really.

kti-cameron: Atws doesn't make `self.at_api_client.picklist['Task']['Status']` available by this method. I guess because the statuses are the same. I think its ok to use this method here since the statuses are the same and we don't have any reason to think that might change soon. Still, we can access the picklist values directly from the Task if we need to, just not using this method.

djautotask/admin.py
kti-cameron: Yeah, very useful. I've added status to the list display, search field and added it as a list_filter.

I also found another issue while cleaning these up. I forgot that order is important for Method Resolution Order. So for `TaskSynchronizer(Synchronizer, QueryConditionMixin,)` the mixin override of _get_query_conditions() was not actually running and wasn't filtering out `Complete` tasks. I've reversed the order to `TaskSynchronizer(QueryConditionMixin, Synchronizer` now so complete tasks and tickets are not synced again.

djautotask/sync.py
kti-sam: Great! That makes sense and seems like this is the intentional way then!

PR #34, opened 2019-11-27
Format API errors better

kti-sam:


djautotask/management/commands/atsync.py
```diff
@@ -106,23 +108,28 @@ def handle(self, *args, **options):
...
             except SAXParseException as e:
-                msg = 'Failed to connect to Autotask API. ' \
+                error_msg = 'Failed to connect to Autotask API. ' \
                       'The error was: {}'.format(e)
 
-                error_messages += '{}\n'.format(msg)
-                failed_classes += 1
+            finally:
+                if error_msg:
+                    self.stderr.write(error_msg)
```
kti-cameron: I think this is unnecessary since the error message gets displayed when the `CommandError` is raised at line 141.

kti-sam: I added it so it works the same as cwsync, rather than having both these methods work and report in different ways.

kti-cameron: Ok, no problem. It just bothered me that in duplicated error messages after a failed sync. Not a big deal though. I'll just run through some final testing and then merge.

PR #36, opened 2019-12-02
1187 add autotask project task boards

kti-cameron:


djautotask/api.py
```diff
@@ -168,20 +168,23 @@ def send(self, request):
...
-    query = Query('Ticket')
-    query.WHERE('id', query.Equals, ticket.id)
-    at = init_api_connection()
-
-    t = at.query(query).fetch_one()
-    t.Status = status.id
-
-    # Fetch one executes the update and returns the created object.
-    return at.update([t]).fetch_one()
+class AutotaskAPIClient(object):
```
kti-sam: This doesn't need to be an object, or if there is a reason for this I'm missing it needs to be renamed. While it does have access to the api client, it is not itself the api client. That is mostly a nitpick I suppose. The more important reason I don't see a need for it is that this `update_object` method gains no benefit from being an on an `AutotaskAPIClient` instance. It has no references to `self`. You could make it a static method, but then there is no reason for it to be on a class with only this method. It may as well be a function on its own again. 

There might be a benefit to making this an class of its own in the future. Maybe having `init_api_connection` in its __init__, having most of the other methods using the connection in the class, and the synchronizers creating an `AutotaskAPIClient` to use those methods. But I don't see a benefit with just this one static method, and more importantly I don't see a benefit in re-factoring `api.py` for this issue.

kti-sam: To clarify, when I say "it is not itself the api client", I mean it doesn't have a reference to the client, it only has one method to interact with AT so it doesn't really facilitate much, etc.

kti-cameron: Yes that makes sense. Might be needed in the future but not now. I originally added it thinking I would need different methods for updating tasks and tickets. Since tasks and tickets are so similar I didn't have to add as many changes as I expected and should've just removed the class then. 

Thanks for catching. Should be fixed now.

PR #37, opened 2019-12-02
v 0.0.31

kti-cameron:


kti-cameron:
Forgot that #35 issue needed its version updated to 31. Just adding that here.

PR #41, opened 2019-12-11
Add project status filter to sync

kti-cameron:
This is more complicated than I'd like but I don't believe that we can query the API to check for inactive project statuses when syncing Tasks or Projects. Thats the reasoning behind adding the `fetch_records` method which determines whether or not to persist the task or project based on its associated project status.


djautotask/sync.py
```diff
@@ -476,9 +481,31 @@ def _assign_field_data(self, instance, object_data):
...
+                    )
+                    continue
+
+            self.persist_record(record, results)
+
+
+class ProjectSynchronizer(FilterProjectStatusMixin, Synchronizer):
     model_class = models.Project
     last_updated_field = 'LastActivityDateTime'
+    lookup_field = 'Status'
```
kti-sam: Lookup field is a little to generic for this I think. I suggest changing this to relate to filtering, or something else that makes sense. Just a little more specific.

djautotask/sync.py
```diff
@@ -7,6 +7,7 @@
 from django.utils import timezone
 
 from djautotask import api, models
+from djautotask.models import PROJECT_STATUS_COMPLETE
```
kti-sam: You are already importing models, so you can just access this with `models.PROJECT_STATUS_COMPLETE`

kti-sam: Actually, this would probably be better included in the `ProjectStatus` class. Then, as a class attribute, you could access it like `ProjectStatus.COMPLETE`, which looks a lot cleaner to me.

djautotask/sync.py
kti-cameron: Fixed.

djautotask/sync.py
kti-cameron: Nice idea. Fixed

PR #43, opened 2019-12-17
1263 add assign null relation method to parent synchronizer class

kti-cameron:


kti-sam:
Unit test for this one would be good

PR #45, opened 2020-01-17
1174 sync autotask time entries

kti-cameron:


djautotask/sync.py
```diff
@@ -673,3 +673,67 @@ def _assign_field_data(self, instance, object_data):
...
+                )
+
+    def persist_time_entry(self, record, results, entity,
+                           object_id, object_ids):
+        log_message = '{} {} is not in the local database. ' \
+                      'Skipping Time Entry: {}.'
+
+        # Only save time entries for tickets or tasks that are
+        # already in the DB
+        if object_ids and object_id not in object_ids:
```
kti-sam: We are doing a lot of checks after the fact to see if these are time entries we care about. We kind of had to do something similar with the project/complete thing because there was not a better alternative. 

Would it be feasible to have a similar method to the `batch_condition_list`s? Rather than requesting ALL time entries, and chucking out the garbage, can we request time entries only for objects we care about? What is the upper limit of ORs we can have on a query? If it is close to the batch limit in CW then we can just do something similar by looping through and adding a bunch. If its a low number then perhaps not, as just this would eat into our query quota.

If the above doesn't work at all, you can make an improvement to remove your if...else block on line 720. is that you could request all time entries where the Ticket field is not null, then we know we are only getting ticket time entries back. And vice versa for tasks. This way will stop us from getting warning spam if they suddenly make a new type of time entry we don't care about, or allow time entries without tickets/tasks.

djautotask/admin.py
```diff
@@ -143,7 +143,7 @@ def get_queryset(self, request):
 class TaskAdmin(admin.ModelAdmin):
     list_display = ('id', 'title', 'number', 'status', 'phase')
     list_filter = ('status', 'phase')
-    search_fields = ('id', 'title', 'number', 'status', 'phase')
```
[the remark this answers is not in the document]
kti-cameron: Yeah I think its the same one. I noticed it shortly before you opened that issue. I'm pretty sure its an issue with the searching status and phase being foreign keys. I just popped this in to see if it worked. I meant to change remove it. I'll remove it now and let it be handled in #1303

djautotask/sync.py
kti-cameron: Yeah I agree. I spent a fair bit of time trying to figure out if we could make something like batch condition list work. I haven't been able to come up with a solution for that though. From what I can see the AT queryxml doesn't have a method like `contains` that the batch condition list uses. On the CW, the batch condition list can just create a list of ticket ids and say to the API, only give me tickets that are in this set. I'm pretty sure the AT API doesn't support this. At least I can't see it at the moment. 

As for the upper limit on ORs. I'm not sure how we could find that out. But I'm not sure it would be practical (if it is possible) to generate a large query of 
```python 
query.WHERE('TicketID, query.Equals, '1')
query.open_bracket('OR')
query.WHERE('TicketID, query.Equals, '2')
```
Is generating a query like this along the lines of what you had in mind? Cause I can't see how else to do it at the moment. 

Good idea with querying for NotNull time entries. I didn't think of that. Now I can remove that logger. 

https://ww4.autotask.net/help/Content/AdminSetup/2ExtensionsIntegrations/APIs/WSAPI/APICalls/QueryXML.htm?Highlight=queryxml

kti-sam: Yes, that is basically what I had in mind, except a loop through X number of ticket ids, where X is the maximum amount of times we can add an open bracket before it takes no more.
```
For id in Ticket.objects.values_list('id', flat=True):
    query.open_bracket('OR')
    query.WHERE('TicketID, query.Equals, '1')
    query.close_bracket()
```

Again, I don't know if we can do that, but it should be pretty easy to find out. We are doing essentially the same thing in CW, cramming about 480 ticket ids in the url for syncing time entries per batch. If you can do that open bracket a comparable amount of times then we should be good to go. If we can only do it something like 50 times though it might not be for the best, as we have the request limit to worry about.

kti-cameron: I did some digging and tried out your suggestion. I used the for loop you like you suggested. 
```python
        for ticket_id in ticket_ids:
            query.open_bracket('OR')
            query.WHERE('TicketID', query.Equals, ticket_id)
            query.close_bracket()
```
When using a `ticket_ids` queryset of over 472 tickets I got the following error message:
```
atws.wrapper.AutotaskAPIException: 'Autotask API has returned an error.\n    Please process the re
sponse attribute in this exception to determine\n    the extent of this failure.\n    '[Some part 
of your SQL statement is nested too deeply. Rewrite the query or break it up into smaller queries.
```
Ticket querysets of less than 472 were successful. So I suppose that means there are 472 of these
`<condition operator="OR"><field>TicketID<expression op="Equals">9624</expression></field></condition>` condition elements in our XML.

PR #47, opened 2020-01-20
fixed search of status and phase on tasks

kti-sam:


kti-cameron:
Thanks for fixing! Looks good. 

Might be out of scope but can I suggest we add in 
```python
    def get_queryset(self, request):
        qs = super().get_queryset(request)
        return qs.select_related('phase', 'status')
```
to fix a bunch of duplicate queries and make the page load a little faster too? If too out of scope let me know and I'll make an issue.

kti-sam:
Good to go

PR #50, opened 2020-01-29
1175 sync autotask notes

kti-sam:


djautotask/sync.py
```diff
@@ -340,6 +340,66 @@ def _get_query_conditions(self, query):
...
+    def get(self, results):
+
+        sync_job_qset = models.SyncJob.objects.filter(
+            entity_name=self.model_class.__name__
+        )
+        object_queries = self.build_batch_queries(sync_job_qset)
+
+        # Apply extra conditions if they exist, else nothing happens
+        for query in object_queries:
+            self._get_query_conditions(query)
```
kti-sam: I just realized this can be moved into the loop with fetch records, just before the fetch. I'll do that soon, still ready for review otherwise.

djautotask/sync.py
```diff
@@ -703,7 +769,78 @@ def _assign_field_data(self, instance, object_data):
...
+            'NoteType',
+            query.NotEqual,
+            self.at_api_client.picklist
+            ['TicketNote']['NoteType']['Workflow Rule Note - Task']
+        )
+        query.close_bracket()
+        return query
+
+
+class TicketNoteSynchronizer(BatchQueryMixin, NoteQueryMixin, Synchronizer):
```
kti-cameron: Most of `TicketNoteSynchronizer` and `TaskNoteSynchronizer` are nearly identical. I think it would make sense to at least move `_assign_field_data` to a parent mixin.

kti-sam: Good to go!

PR #51, opened 2020-01-30
1206 display time entries on autotask cards like in cw kanban

kti-cameron:


djautotask/models.py
```diff
@@ -1,8 +1,11 @@
 from django.db import models
 from django.db.models import Q
 from django_extensions.db.models import TimeStampedModel
+from django.utils import timezone
 from djautotask import api
 
+OFFSET_TIMEZONE = 'America/New_York'
```
kti-sam: Since we have this in two files now, lets import it from a single source. Don't put it in the callable though, its never going to change based on who is using this.

kti-cameron: Yeah makes sense. I'm thinking we keep `OFFSET_TIMEZONE` here in models.py and then import it to use in taskboard.py. Unless you've got somewhere else in mind.

PR #55, opened 2020-02-05
added role sync

kti-sam:


djautotask/models.py
```diff
@@ -520,3 +523,18 @@ def get_entered_time(self):
             entered_time = date_worked - local_offset
 
         return entered_time
+
+
+class Role(models.Model):
+    name = models.TextField(blank=True, null=True, max_length=200)
```
kti-cameron: Should this be a CharField instead since its max length is only 200? Same goes for description.

djautotask/models.py
```diff
@@ -520,3 +523,18 @@ def get_entered_time(self):
...
+    active = models.BooleanField(default=True)
+    description = models.TextField(blank=True, null=True, max_length=200)
+    hourly_factor = models.DecimalField(
+        blank=True, null=True, decimal_places=2, max_digits=9)
+    hourly_rate = models.DecimalField(
+        blank=True, null=True, decimal_places=2, max_digits=9)
+    role_type = models.PositiveIntegerField(blank=True, null=True)
+    system_role = models.BooleanField(default=False)
+
+    def __str__(self):
```
kti-cameron: The API docs say that name is a required field. I don't really have a problem with allowing null names but if we're going to do that we should probably check if `self.name` is null here.

kti-cameron:
Thanks for the fixes. Merging

PR #58, opened 2020-02-06
added remaining role entity syncs

kti-sam:


djautotask/models.py
```diff
@@ -565,3 +565,35 @@ class Department(models.Model):
 
     def __str__(self):
         return self.name
+
+
+class ResourceRoleDepartment(models.Model):
```
kti-cameron: I just noticed that these are not `TimeStampedModel`'s like most of the other models. Probably doesn't really matter but any reason why?

kti-sam: Do these models need any of the methods or attributes of a `TimeStampedModel`? I didn't think they did, so I did not include it. I thought the those were used for dealing with partial syncs and querying in CW and AT or something, so we didn't need them on models with no `last_updated_field`. Let me know if I am wrong about that.

kti-cameron: All `TimeStampedModel` provides is created and modified fields. I could be mistaken but I don't think we actually use the created or modified fields in the synchronizers. And the `last_updated_field` is a field on the model so is not related to the created or modified fields. 

Anyway, I figured it was worth asking. Seeing as we don't appear to really need them right now. We can always add it later if we do need it.

kti-cameron:
Looks good. Merging

PR #60, opened 2020-02-20
1210 with default resoure roles

kti-cameron:


djautotask/sync.py
```diff
@@ -416,7 +416,8 @@ class TicketSynchronizer(QueryConditionMixin, Synchronizer):
         'Source': (models.Source, 'source'),
         'IssueType': (models.IssueType, 'issue_type'),
         'SubIssueType': (models.SubIssueType, 'sub_issue_type'),
-        'AssignedResourceRoleID': (models.Role, 'assigned_resource_role'),
+        'AssignedResourceRoleID': (models.Role, 'role'),
```
kti-sam: This was left in after removing role

djautotask/sync.py
```diff
@@ -737,7 +739,8 @@ class TaskSynchronizer(QueryConditionMixin,
         'PhaseID': (models.Phase, 'phase'),
         'Status': (models.Status, 'status'),
         'PriorityLabel': (models.Priority, 'priority'),
-        'AssignedResourceRoleID': (models.Role, 'assigned_resource_role'),
+        'AssignedResourceRoleID': (models.Role, 'role'),
```
kti-sam: Same as above

djautotask/sync.py
kti-cameron: Oops. Removed the wrong line. Fixed now.

PR #63, opened 2020-03-04
initial_id improvements

kti-sam:


djautotask/sync.py
```diff
@@ -73,6 +73,7 @@ def __init__(self):
 
 class Synchronizer:
     lookup_key = 'id'
+    db_lookup_key = lookup_key
```
kti-cameron: I don't see why we can't just use `lookup_key` here?

kti-cameron: Discussed in person. Thanks for clearing that up.

kti-sam: Explained in person, but for posterity, lookup_key is already in use by Picklist synchronizers for api instance lookups, and the lookup for its api instance form and record form is different.

PR #73, opened 2020-04-03
added note creation

kti-sam:


djautotask/models.py
```diff
@@ -302,35 +305,49 @@ def __str__(self):
         return '{} {}'.format(self.resource, self.ticket)
 
 
-class TicketNote(TimeStampedModel):
+class Note:
+    ALL_USERS = 1
+    INTERNAL_USERS = 2
+    PUBLISH_CHOICES = ((ALL_USERS, 'All Autotask Users'),
+                       (INTERNAL_USERS, 'Internal Project Team'))
```
kti-cameron, line 312: Should `Internal Project Team` be `Internal Users` instead? I haven't seen any mention of `Internal Project Team` in the AT UI.

kti-sam: Ah oops, good catch. I pulled that from the API actually. Thats what the picklist field reported it as.

kti-cameron: Oh yeah I see that now. That confusing

kti-sam: I changed it anyway so it makes more sense to Autotask users, since its "Internal Project Team" even for Tickets.

PR #74, opened 2020-04-15
sync service call models

kti-sam:


djautotask/admin.py
```diff
@@ -257,3 +257,58 @@ class ContractAdmin(admin.ModelAdmin):
 
     def status_name(self, obj):
         return obj.STATUS_CHOICES[int(obj.status)][1]
+
```
kti-cameron, line 260: It looks like we're missing a model admin for `ServiceCallStatus`

PR #97, opened 2020-09-18
1575 sync taskpredecessor records

kti-cameron:


djautotask/models.py
```diff
@@ -871,3 +871,17 @@ class ServiceCallTaskResource(TimeStampedModel):
...
+
+
+class TaskPredecessor(TimeStampedModel):
+    lag_days = models.IntegerField(blank=True, null=True)
+    predecessor_task = models.ForeignKey('Task', blank=True, null=True,
+                                         related_name='predecessor_task',
+                                         on_delete=models.CASCADE)
+    successor_task = models.ForeignKey('Task', blank=True, null=True,
+                                       related_name='successor_task',
+                                       on_delete=models.CASCADE)
```
kti-sam, line 883: This will delete tasks when other tasks are deleted from the DB in syncs, if I am understanding correctly.

kti-cameron: I was under the impression that is was the other way around. i.e. a task is deleted and the associated `TaskPredecessor` is also deleted. But FK task wouldn't get deleted if the `TaskPredecessor` is deleted. 


But yeah, I agree SET_NULL would be better in case the predecessor task gets removed but not the successor or the other way around. I'll change to SET_NULL

kti-sam: No sorry, I was very wrong. I somehow thought those were fields on `Task`, not the obvious TaskPredecessor model, Cascade is the best option because is we don't have any reference to the predecessor we can't get any information about it to give to a user anyway. I guess I was still a little sick when I reviewed this. I'll revert that commit and merge soon.

PR #100, opened 2020-10-01
Add task predecessor batch fix

kti-cameron:


kti-sam:
Made one small best practice commit. Its best to define attributes on the class or in __init__, rather than other methods like `setUp`. `setUp` is best for initializing instance specific things, or setting up database test data.

PR #108, opened 2020-12-14
ADD: Contact sync REST API

kti-daniel:


djautotask/models.py
```diff
@@ -372,6 +375,32 @@ class TaskNote(TimeStampedModel, Note):
     )
 
 
+class Contact(TimeStampedModel):
+
+    first_name = models.TextField(blank=True, null=True, max_length=20)
+    last_name = models.TextField(blank=True, null=True, max_length=20)
+    middle_initial = models.TextField(blank=True, null=True, max_length=50)
```
kti-cameron, line 382: Since we are syncing a minimal amount of fields for CW contacts, I think it's reasonable to do the same here. See comment at https://github.com/KerkhoffTechnologies/django-connectwise/pull/308#discussion_r542930266 . 

I think everything can be removed except: first_name, last_name, email_address and account.

djautotask/syncrest.py
```diff
@@ -0,0 +1,414 @@
...
+
+    related_meta = {
+        'companyID': (models.Account, 'account'),
+    }
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.api_conditions = ['inactiveFlag=False']
+
+    def _assign_field_data(self, instance, json_data):
```
kti-cameron, line 388: Right now whenever `atsyncrest` is run any records that are returned are saved to the database. What we want is to run a sync and for each record returned check our database to see if we have the record, if we have it then update it, and if not save it as a new record.

We need to make sure that we set the `id` on the instance to the `id` that we get from the AT API. Something like `instance.id = json_data.get('id')`. This will only the synchronizer's to be able to find out if the record from the API is in our database.

djautotask/syncrest.py
```diff
@@ -0,0 +1,414 @@
...
+        instance.first_name = json_data.get('firstName')
+        instance.last_name = json_data.get('lastName')
+        instance.middle_initial = json_data.get('middleInitial')
+        instance.additional_address_information = json_data.\
```
kti-cameron, line 392: As mentioned on the model code, most of these fields can be removed.

djautotask/migrations/0087_auto_20201221_1526.py
```diff
@@ -0,0 +1,57 @@
+# Generated by Django 3.1.2 on 2020-12-21 15:26
```
kti-cameron, line 1: It would be good to delete these 2 migration files and make a single new migration for the changes. As it currently is, the first migration adds a number of fields and the next migration removes some that were just added.

kti-cameron:
Overall this looks really good @kti-daniel . I still need to go over some more parts of it in detail.

The two biggest changes that are needed are:
1. adding pagination for the results from the API
2. adding ability to run partial syncs by setting an API conditions to query for records that have been updated since a given time.

We need pagination because the API only returns a maximum of 500 records per request. We need to be able to handle getting more than 500 records. For adding pagination, it looks like AT makes it pretty easy since it looks like requests contain
```
 "pageDetails": {
        "count": 41,
        "requestCount": 500,
        "prevPageUrl": null,
        "nextPageUrl": null
    }
```
So you should be able to put something in your `fetch_records` method to check if there is a `nextPageUrl`. If it is present then use it to make the next request. Could be quite similar to the while loop in django-connectwise `fetch_records` method.

More details on pagination at https://ww4.autotask.net/help/DeveloperHelp/Content/AdminSetup/2ExtensionsIntegrations/APIs/REST/API_Calls/REST_API_Calls.htm


At the moment, partial syncs are not working for syncrest.py. You'll have to adjust it by making sure that if the --full option is not present then set conditions to make sure to only fetch records from the API since the last sync was run. The field you'll want to query on is probably `lastActivityDate` for Contact. lastUpdated is only a field in CW.

djautotask/syncrest.py
```diff
@@ -0,0 +1,402 @@
...
+            initial_ids = self._instance_ids()
+
+            results.deleted_count = self.prune_stale_records(
+                initial_ids, results.synced_ids
+            )
+
+        return results.created_count, results.updated_count, \
+            results.skipped_count, results.deleted_count
+
+    def callback_sync(self, filter_params):
```
kti-cameron, line 333: I think this method can be removed for now. We will probably need something similar later on when we move handling of webhook/callbacks to the REST API but no need to add now.

djautotask/api_rest.py
```diff
@@ -0,0 +1,359 @@
...
+
+from djautotask.utils import DjautotaskSettings
+
+CW_RESPONSE_MAX_RECORDS = 500  # The greatest number of records Autotask
+# will send us in one response.
+RETRY_WAIT_EXPONENTIAL_MULTAPPLIER = 1000  # Initial number of milliseconds to
+# wait before retrying a request.
+RETRY_WAIT_EXPONENTIAL_MAX = 10000  # Maximum number of milliseconds to wait
+# before retrying a request.
+AT_DEFAULT_PAGE = 1  # CW Pagination is 1-indexed
```
kti-cameron, line 18: I don't think this is being needed or being used.

djautotask/api_rest.py
```diff
@@ -0,0 +1,359 @@
...
+CONTENT_DISPOSITION_RE = re.compile(
```
kti-cameron, line 19: This can be removed too

djautotask/api_rest.py
```diff
@@ -0,0 +1,359 @@
...
+
+class ContactsAPIClient(AutotaskAPIClient):
+    API = 'contacts'
+
+    def by_id(self, contact_id):
+        endpoint_url = '{}'.format(contact_id)
+        return self.fetch_resource(endpoint_url)
+
+    def get_contacts(self, *args, **kwargs):
+        # TODO
```
kti-cameron, line 353: Can this be removed?

djautotask/models.py
```diff
@@ -372,6 +375,20 @@ class TaskNote(TimeStampedModel, Note):
     )
 
 
+class Contact(TimeStampedModel):
+
+    first_name = models.TextField(blank=True, null=True, max_length=20)
```
kti-cameron, line 380: Since the first_name and last_name fields only have a max length of 20 these fields should be `CharField`s https://docs.djangoproject.com/en/3.1/ref/models/fields/#charfield

kti-cameron:
You should add some simple tests for the Contact synchronizer and atsync contact management command since it will likely be quite different from the tests for sync classes using the SOAP API.

djautotask/syncrest.py
kti-daniel: Hi Cameron, regarding pagination in AT RESTAPI, it is a little different with that in CW. In CW, we know the exact address of the 2nd page, and we can call it directly, which is the same as access array. But in AT, it is something like iterator (bi-directional). i.e. we don't know the address of 2nd page of the API. When we got the result of the API, it contains the addresses in neighbourhood such as previous url and next url.

kti-daniel: It looks something like:

    "pageDetails": {
        "count": 3,
        "requestCount": 3,
        "prevPageUrl": null,
        "nextPageUrl": "https://webservices2.autotask.net/atservicesrest/v1.0/Contacts/query/next?paging=%7b%22pageSize%22%3a3%2c%22previousIds%22%3a%5b29683585%5d%2c%22nextIds%22%3a%5b29683588%5d%7d"
    }

Therefore, it has no direct access feature of some page, and it has different structure as it is.

djautotask/migrations/0087_auto_20201221_1526.py
kti-daniel: Done. Thanks~ :+1:

djautotask/syncrest.py
kti-daniel: I commented too rapidly after reading "1. adding pagination for the results from the API".

Never mind the previous comments. Anyway, it has pagination feature and it can handle the maximum number of records, 500. It seems a good place to get nextPageUrl during parsing API results, and it locates in 
_fetch_resource method.

            if 200 <= response.status_code < 300:
                rslt_items = response.json()
                return_items.append(rslt_items.get("items"))
                next_url = rslt_items.get("pageDetails").get("nextPageUrl")

                if next_url:
                    retry_counter['count'] = 0
                    _fetch_resource(next_url, full_next_url=True,
                                    retry_counter=retry_counter,
                                    *args, **kwargs)

djautotask/api_rest.py
kti-daniel: Good point. RETRY_WAIT_EXPONENTIAL_MAX is used in fetch_resource, but AT_DEFAULT_PAGE is not. Thanks

djautotask/api_rest.py
kti-daniel: Done

djautotask/models.py
kti-daniel: Excellent point.

kti-daniel:
Some findings needed to be shared during supporting partial sync in AT:

In CW, _info fields have the same lastUpdateDate info per response(1 lastUpdateDate/ 1 response), but each item has each lastActivityDate info in AT(multiple lastActivityDates/ 1 response).

In partial sync in AT, we can filter using lastActivityDate, and it has benefit to reduce network traffic.

What I found is that GET & POST work differently even if AT manual says it works the same.

the following works with POST and returns data, but

{
    "filter": [
        {
            "op": "eq",
            "field": "IsActive",
            "value": true
        },
        {
            "op": "gt",
            "field": "lastActivityDate",
            "value": "2020-12-21T16:53:47.937025+00:00"
        }
    ]
}

but, the GET requests (both not-encoded or encoded) don't work with the same data above.

https://webservices2.autotask.net/ATServicesRest/v1.0/contacts/query?search=%7B%22filter%22:%20%5B%7B%22op%22:%20%22eq%22,%20%22field%22:%20%22IsActive%22,%20%22value%22:%20%22true%22%7D,%20%7B%22op%22:%20%22gt%22,%20%22field%22:%20%22lastActivityDate%22,%20%22value%22:%20%222020-12-21T16:53:47.937025+00:00%22%7D%5D%7D   (in encoded case, total 320 characters)

with the below error
"No message. Conversion failed when converting date and/or time from character string." (w/ code 500)

One more thing to be changed is how to find lastActivityDate, which is last success partial sync date:

        sync_job_qset = self.get_sync_job_qset().filter(success=True)

        if sync_job_qset.count() > 1 and not self.full:
            last_sync_job_time = sync_job_qset.last().start_time.isoformat()
            self.api_conditions.append(
                "lastActivityDate,{0},gt".format(last_sync_job_time)
            )

djautotask/api_rest.py
```diff
@@ -0,0 +1,359 @@
...
+import re
+
+import requests
+from django.conf import settings
+from retrying import retry
+from django.core.cache import cache
+
+from djautotask.utils import DjautotaskSettings
+
+CW_RESPONSE_MAX_RECORDS = 500  # The greatest number of records Autotask
```
kti-cameron, line 12: This should be AT_RESPONSE_MAX_RECORDS. But is it being used anywhere? If not it can be removed.

kti-daniel:
> even if AT manual says it works the same.

One potential problem found:
the nextPageUrl doesn't work because AT provides us with GET address, so using POST method could not be the final solution here.

    "pageDetails": {
        "count": 3,
        "requestCount": 3,
        "prevPageUrl": null,
        "nextPageUrl": "https://webservices2.autotask.net/ATServicesRest/v1.0/contacts/query/next?paging=%7b%22pageSize%22%3a3%2c%22previousIds%22%3a%5b29683585%5d%2c%22nextIds%22%3a%5b30683993%5d%7d&search=%7b%22filter%22%3a+%5b%7b%22op%22%3a+%22eq%22%2c+%22field%22%3a+%22IsActive%22%2c+%22value%22%3a+%22true%22%7d%2c+%7b%22op%22%3a+%22gt%22%2c+%22field%22%3a+%22lastActivityDate%22%2c+%22value%22%3a+%222020-12-21T16%3a53%3a47.937025+00%3a00%22%7d%5d%7d"
    }

djautotask/api_rest.py
```diff
@@ -0,0 +1,360 @@
...
+
+            try:
+                if full_next_url:
+                    endpoint = endpoint_url
+                else:
+                    endpoint = self._endpoint(self.POST_QUERY)
+
+                logger.debug('Making POST request to {} with {}'.format(
+                    endpoint, payload))
+                response = requests.post(
```
kti-cameron, line 257: Replying here because GitHub doesn't allow me to reply to your note about GET requests not working when querying with `lastActivityDate`. 

I think it would be preferred to keep this a GET request if possible. I just tested a similar request in Postman successfully with 
```
https://webservices2.autotask.net/atservicesrest/v1.0/contacts/query/?search={"filter": [{"op": "eq", "field": "IsActive", "value": "true"}, {"op": "gt", "field": "lastActivityDate", "value": "2019-11-26T18:12:18.567Z"}]}
```
I formatted the date using `isoformat` which is what django-connectwise does to format it's lastUpdated date before sending to the API. I think this is why you were getting the "No message. Conversion failed when converting date and/or time from character string." error. 

So I suggest changing back to a GET request and making sure that the date you use in your request is iso formatted.

kti-cameron: I was wrong. I see now that the date is already getting formatted with `isoformat`. But it appears that the API has the conversion error when passing a date string in this format: "2020-12-23T20:42:23.224615+00:00". 

I don't see the error if I use a string in this format: "2020-12-23T20:42:23.224615Z". So maybe this is the way to go. Give that a try and let me know if you are still seeing problems. 

To get that date formatted you should be able to use something like:
```
        if sync_job_qset.count() > 1 and not self.full:
            last_sync_job_time = sync_job_qset.last().start_time.strftime('%Y-%m-%d %H:%M:%S%fZ')
            self.api_conditions.append(
                "lastActivityDate,{0},gt".format(last_sync_job_time)
            )
```

djautotask/api_rest.py
```diff
@@ -0,0 +1,360 @@
...
+                )
+
+            except requests.RequestException as e:
+                logger.error('Request failed: POST {} with {}: {}'.format(
+                    endpoint, payload, e))
+                raise AutotaskAPIError('{}'.format(e))
+
+            if 200 <= response.status_code < 300:
+                rslt_items = response.json()
+                return_items.append(rslt_items.get("items"))
```
kti-cameron, line 271: I'd prefer to move the code that decides if there is another page and request to make out of the `_fetch_resource` method. I think it would be simpler and easier to understand to have something that is more similar to the `fetch_records` method in django-connectwise. 

Another thing that concerns me with your current implementation is that it looks like it keeps appending records to the `return_items` list for as long as records are returned from the API. I suspect this could be a problem when run with a large number of records. 

The django-connectwise `fetch_records` method doesn't have this problem since it fetches one page of records from the API, then persists the records in the TopLeft DB, and then fetches another page of records. 

So more along the lines of:
```
        while True:
            logger.info(
                'Fetching {} records'.format(
                    self.model_class.__bases__[0].__name__)
            )
            page = self.get_page()
            self.persist_page(page, results)

            next_url = page.get("pageDetails").get("nextPageUrl")
            if not next_url:
                break
```
Of course the `get_page` method would have to accept next_url somehow for `fetch_resource` to use later. Let me know if you see any problems with this approach.

djautotask/api_rest.py
kti-sam: As discussed in Zoom call, I agree with Cameron on this point.

djautotask/api_rest.py
```diff
@@ -0,0 +1,342 @@
...
+            else:
+                self._log_failed(response)
+                raise AutotaskAPIError(
+                    self._prepare_error_response(response))
+
+        # PAYLOAD should be the same through the all the requests of any page
+        if not self.PAYLOAD:
+            self.build_payload(**kwargs)
+
+        next_url = args[0]
```
kti-sam, line 284: As discussed in call, this can be changed to a keyword argument for better readability and maintainability.

djautotask/sync_rest.py
```diff
@@ -0,0 +1,367 @@
...
+logger = logging.getLogger(__name__)
+
+
+class InvalidObjectException(Exception):
+    """
+    If for any reason an object can't be created (for example, it references
+    an unknown foreign object, or is missing a required field), raise this
+    so that the synchronizer can catch it and continue with other records.
+    """
+    pass
```
kti-sam, line 22: This is a redefinition from sync.py, you can import and use that instead.

djautotask/sync_rest.py
```diff
@@ -0,0 +1,367 @@
...
+    lookup_key = 'id'
+
+    def __init__(self, full=False, *args, **kwargs):
+        self.api_conditions = []
+        self.client = self.client_class()
+        self.full = full
+
+        self.pre_delete_callback = kwargs.pop('pre_delete_callback', None)
+        self.pre_delete_args = kwargs.pop('pre_delete_args', None)
+        self.post_delete_callback = kwargs.pop('post_delete_callback', None)
```
kti-sam, line 82: These three lines about callbacks can be removed

djautotask/sync_rest.py
```diff
@@ -0,0 +1,367 @@
...
+                self.model_class.objects.filter(pk=instance_id).delete()
+            finally:
+                if post_delete_callback:
+                    post_delete_callback(pre_delete_result)
+            logger.info(
+                'Deleted {} {} (if it existed).'.format(
+                    self.model_class.__bases__[0].__name__,
+                    instance_id
+                )
+            )
```
kti-sam, line 224: As per call, remove `fetch_sync_by_id` and `fetch_delete_by_id`

djautotask/sync_rest.py
```diff
@@ -0,0 +1,367 @@
...
+                    *self.pre_delete_args
+                )
+            logger.info(
+                'Removing {} stale records for model: {}'.format(
+                    len(stale_ids), self.model_class.__bases__[0].__name__,
+                )
+            )
+            delete_qset.delete()
+            if self.post_delete_callback:
+                self.post_delete_callback(pre_delete_result)
```
kti-sam, line 300: Remove code here related to `pre_delete_result`, `pre_delete_callback`, and `post_delete_callback`

djautotask/sync_rest.py
kti-sam: Also don't duplicate `log_sync_job`, or `SyncResults`

kti-daniel:
Done. Thanks Sam.

kti-daniel:
Completed. Thanks again.

djautotask/tests/test_model.py
```diff
@@ -51,34 +51,38 @@ def assert_get_entered_time_date_worked(self, local_midnight):
...
+    #
+    #     self.assert_get_entered_time_date_worked(local_midnight)
+    #
+    # @override_settings(TIME_ZONE='Europe/Paris')
+    # def test_get_entered_time_date_worked_paris(self):
+    #
+    #     local_midnight = timezone.localtime().replace(
+    #         hour=0, minute=0, second=0, microsecond=0)
+    #
+    #     self.assert_get_entered_time_date_worked(local_midnight)
```
kti-sam, line 88: I'm un-commenting these now and committing before I merge, as they only break on new years eve, every other day they are fine.

PR #110, opened 2021-01-06
Fix django autotask unit tests that break on new years eve

kti-daniel:


djautotask/models.py
```diff
@@ -380,6 +380,7 @@ class Contact(TimeStampedModel):
     first_name = models.CharField(blank=True, null=True, max_length=20)
     last_name = models.CharField(blank=True, null=True, max_length=20)
     email_address = models.CharField(blank=True, null=True, max_length=50)
+    phone = models.CharField(blank=True, null=True, max_length=25)
```
kti-sam, line 383: It looks like you have accidentally added an extra commit in this pull request. There is some extra stuff here about contacts. Please fix this branch and pull request to just the relevant changes.

kti-sam:
I fixed the merge conflict, but I found another problem. Should be easy enough to fix.

kti-sam:
There was an unused import blocking unit tests so I pushed a commit removing that. Merging shortly.

PR #119, opened 2021-03-08
MOD: Time entry show the last modified time as end date when the end …

kti-daniel:
…date of task cards is null


djautotask/models.py
```diff
@@ -664,21 +664,13 @@ def get_entered_time(self):
...
-            est_offset = timezone.localtime(
-                timezone=timezone.pytz.timezone(OFFSET_TIMEZONE)).utcoffset()
-            local_offset = timezone.localtime().utcoffset()
-
-            # We want to end up with a UTC datetime that is midnight in the
-            # local timezone.
-            date_worked = self.date_worked + est_offset
-            entered_time = date_worked - local_offset
+            # record modified time is used in the absence of end_date_time
+            entered_time = self.modified
```
kti-sam, line 673: This doesn't work. The comments explain why we need to use the date_worked field. If you find yourself deleting large sections of commented code please consult myself, Cameron, or Matt. 

By changing this to date modified that will change the time entries date to anytime its data is updated, which does not reflect when a time entry was created.

Looking at the issue description it seems like the original theory for why this bug was ocurring is wrong. So I will put some time into researching the customer's problem again, and reproducing it before you continue work on this issue.

PR #121, opened 2021-04-21
1798 convert autotask ticket sync and api classes to rest

kti-daniel:


djautotask/sync_rest.py
```diff
@@ -271,3 +291,130 @@ def _assign_field_data(self, instance, json_data):
...
+            self._assign_udf_data(instance, udfs)
+
+        if sla_paused:
+            instance.service_level_agreement_paused_next_event_hours = \
+                Decimal(str(round(sla_paused, 2)))
+        if instance.estimated_hours:
+            instance.estimated_hours = \
+                Decimal(str(round(instance.estimated_hours, 2)))
+
+        self.set_relations(instance, json_data)
```
kti-cameron, line 389: I'm seeing block of warning logs for almost every ticket.

```
[2021-04-21 14:15:30] INFO [djautotask.sync_rest:195] Skipped: Ticket 9171-Document the Debian                                               
[2021-04-21 14:15:30] WARNING [djautotask.sync_rest:67] Failed to find projectID None for <class 'djautotask.models.TicketTracker'> 9172.    
[2021-04-21 14:15:30] WARNING [djautotask.sync_rest:67] Failed to find source None for <class 'djautotask.models.TicketTracker'> 9172.       
[2021-04-21 14:15:30] WARNING [djautotask.sync_rest:67] Failed to find issueType None for <class 'djautotask.models.TicketTracker'> 9172.    
[2021-04-21 14:15:30] WARNING [djautotask.sync_rest:67] Failed to find subIssueType None for <class 'djautotask.models.TicketTracker'> 9172. 
[2021-04-21 14:15:30] WARNING [djautotask.sync_rest:67] Failed to find billingCodeID None for <class 'djautotask.models.TicketTracker'> 9172.
[2021-04-21 14:15:30] WARNING [djautotask.sync_rest:67] Failed to find contractID None for <class 'djautotask.models.TicketTracker'> 9172.   
[2021-04-21 14:15:30] WARNING [djautotask.sync_rest:67] Failed to find contactID None for <class 'djautotask.models.TicketTracker'> 9172.    
```

I debugged a little bit and it looks like the API returns all fields for a Ticket entity even if the fields aren't set to anything.  In contrast, CW just doesn't return a field in the API if the field is not set/has no value.

The json_data has a lot of keys with value `None`. So I suspect the code in `_assign_relation` is attempting to lookup a lot of `related_instance = model_class.objects.get(pk=uid)` where uid is None.

djautotask/sync_rest.py
```diff
@@ -271,3 +291,130 @@ def _assign_field_data(self, instance, json_data):
...
+        'contactID': (models.Contact, 'contact'),
+    }
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        request_settings = DjautotaskSettings().get_settings()
+        self.completed_date = (timezone.now() - timezone.timedelta(
+                hours=request_settings.get('keep_completed_hours')
+            )).strftime('%Y-%m-%dT%H:%M:%S.%fZ')
+        self.api_conditions = [
```
kti-cameron, line 326: I think there is something wrong with the API query conditions. Running `./manage.py atsync` should be a partial sync and only return records that have changed since the last sync job but I'm seeing all Ticket records returned on every sync.

djautotask/sync_rest.py
```diff
@@ -129,6 +135,20 @@ def get_single(self, *args, **kwargs):
     def _assign_field_data(self, instance, api_instance):
         raise NotImplementedError
 
+    def _parse_date_time(self, instance, attribute_name):
```
kti-cameron, line 138: I think this method could be simplified and cleaned up a bit. For a start, if the instance that got passed is having an attribute set on it do we really need to return the instance? I would expect the updated attribute to already be set in `_assign_field_data` after setattr is run.

It also seems like this method does more than just parse a datetime - it also sets the attributes on the object.

kti-cameron: Actually do we need this method at all?

I might be missing something obvious but could we do something like:
```python
 first_response_date_time = json_data.get('firstResponseDateTime')
 if first_response_date_time:
     instance.first_response_date_time = parse(first_response_date_time)
```
in the `_assign_field_data` method?

djautotask/api_rest.py
```diff
@@ -335,7 +348,18 @@ def request(self, method, endpoint_url, body=None):
...
 
 class ContactsAPIClient(AutotaskAPIClient):
-    API = 'contacts'
+    API = 'Contacts'
 
     def get_contacts(self, next_url, *args, **kwargs):
         return self.fetch_resource(next_url, *args, **kwargs)
+
+
+class TicketsAPIClient(AutotaskAPIClient):
```
kti-cameron, line 357: I think nearly all other API client classes we have in django-connectwise and django-autotask use singular instead of plural for the class name. So `TicketAPIClient` instead of `TicketsAPIClient`

djautotask/sync_rest.py
```diff
@@ -129,6 +135,20 @@ def get_single(self, *args, **kwargs):
...
+        if getattr(instance, attribute_name):
+            setattr(instance,
+                    attribute_name,
+                    parse(getattr(instance, attribute_name))
+                    )
+        return instance
+
+    def fetch_sync_by_id(self, instance_id):
```
kti-cameron, line 146: This is what I get when calling this method:
```sh
In [8]: ts = TicketSynchronizer()                                                                  
                                                                                                   
In [9]: ts.fetch_sync_by_id(8885)                                                                  
[2021-04-21 15:55:34] DEBUG [djautotask.api_rest:267] Making GET request to Tickets/8885
[2021-04-21 15:55:34] ERROR [djautotask.api_rest:275] Request failed: GET Tickets/8885: Invalid URL
 'Tickets/8885': No schema supplied. Perhaps you meant http://Tickets/8885?                        
```
I think this points to the `get_ticket` method not building the correct URL.

kti-cameron: Since the callouts make use of `fetch_sync_by_id` they are also affected by this error. This is what I get after receiving a ticket callout:
```
djautotask.api_rest.AutotaskAPIError: Invalid URL 'Tickets/8885': No schema supplied. Perhaps you m
eant http://Tickets/8885?
```

djautotask/sync_rest.py
```diff
@@ -271,3 +291,130 @@ def _assign_field_data(self, instance, json_data):
...
+    def get_page(self, next_url=None, *args, **kwargs):
+        kwargs['or_items'] = self.api_conditions
+        return self.client.get_tickets(next_url, *args, **kwargs)
+
+    def get_single(self, ticket_id):
+        return self.client.get_ticket(ticket_id)
+
+    def fetch_sync_by_id(self, instance_id):
+        instance = super().fetch_sync_by_id(instance_id)
+        if instance.completed_date > parse(self.completed_date):
```
kti-cameron, line 401: Is it a replacement for `if instance.status.id != models.Status.COMPLETE_ID:`? If so, I don't see how comparing completed_dates is better than checking for a closed status.

djautotask/sync_rest.py
kti-daniel: Good point! Thanks Cameron.

djautotask/sync_rest.py
kti-daniel: Another good point! Thanks

djautotask/sync_rest.py
kti-daniel: Good point as well! I was a little bit lazy to check if we actually need `return instance`. 

However, I think this method is useful for further usage. Comparing to SOAP API, which returns datetime object, REST_API returns string type, so in a future API method with a datetime info string we need to parse, which could our codes double or more. So I feel some function could be useful at this moment.

I also changed method name to `_set_datetime_attribute`.

Finally, I think we need to parse whether `if first_response_date_time:` is true of not. i.e. it can be set None or some value. Therefore, the feature of function is slightly different from your code snippets. :)

djautotask/api_rest.py
kti-daniel: As I remember, the naming here comes from API documents from AT. It was changed from singular to plural (SOAP to REST), then it was also applied to our REST API. It is now different with ones in CW, but it is the same with API docs in AT. What do you think about that?

image

kti-cameron: ok. If that's what the docs are calling it now then I guess it's fine :man_shrugging:  . Thanks for pointing it out, I didn't realize that.

djautotask/sync_rest.py
kti-cameron: Yeah fair enough. The name change is an improvement and it makes the code look a little cleaner.

djautotask/sync_rest.py
kti-cameron: Thanks for fixing. Callouts work now

djautotask/sync_rest.py
kti-cameron: I still don't understand this. Can you explain it to me a bit?

kti-daniel: Ooops! Is that sitll there? Just a min. It should be changed.... Can't you see the change

image
```
    def fetch_sync_by_id(self, instance_id):
        instance = super().fetch_sync_by_id(instance_id)
        if instance.status.id != models.Status.COMPLETE_ID:
            self.sync_related(instance)
        return instance
```

kti-daniel: Yes, it was changed:

image

kti-cameron: Oops! My mistake. I had the wrong commit selected. Sorry about that. Thanks for clarifying

djautotask/sync_rest.py
```diff
@@ -271,3 +293,134 @@ def _assign_field_data(self, instance, json_data):
     def get_page(self, next_url=None, *args, **kwargs):
         kwargs['conditions'] = self.api_conditions
         return self.client.get_contacts(next_url, *args, **kwargs)
+
+
+# ParentSynchronizer: using SOAP API
+class TicketSynchronizer(SyncRecordUDFMixin, Synchronizer, ParentSynchronizer):
```
kti-cameron, line 299: I think I'm seeing a problem with looking up UDF records. I get

```
[2021-04-24 13:19:55] DEBUG [djautotask.sync:319] No UDF records returned for Name: TicketUDF mat
ching query does not exist.                                                                      
[2021-04-24 13:19:55] DEBUG [djautotask.sync:319] No UDF records returned for Name: TicketUDF mat
ching query does not exist.                                                                      
[2021-04-24 13:19:55] DEBUG [kanban.autotask.models.ticketboard:586] Handling Ticket pre_save sig
nal for 9990.                                                                                    
[2021-04-24 13:19:55] INFO [kanban.autotask.models.ticketboard:565] Handling Ticket post_save sig
nal for 9990 (status is New).                                                                    
[2021-04-24 13:19:55] INFO [djautotask.sync_rest:197] Updated: Ticket 9990-Investigate 404       
Ticket Sync Summary - Created: 0, Updated: 1, Skipped: 0                                         
```
when syncing a ticket with UDF fields set. I'm thinking maybe the REST API stores UDF data differently and the current UDF mixin might need to be adjusted. 


Can you confirm if you are able to sync UDF info on tickets correctly?

kti-daniel: Oh, thanks for the test which I missed. I will look into that.

kti-daniel: https://git.ops.topleft.team/apps/django-topleft/-/issues/1798#note_28544

With permission modified and the most recent code, UDF is now working.

image

Thanks Cameron!

kti-cameron: That's good news! I thought we might've needed a lot more effort to get it working. Thanks for sorting that out Daniel.

kti-daniel: Ooops, I forgot to push!

Please confirm the format is the same as it was stored before. I have no idea and experience about picklist, so I am not sure it is correctly handled. Could you check about that? And could you let me know or show how to test picklist and what it is. Thanks Cemeron.

kti-cameron: No problem. 

UDFs that are picklists look like they are being displayed correctly. Picklists just specify some values/choices that can be chosen for a UDF field. Just like a choice field in a form.  For example, the Skill Level UDF field has a picklist of choices: Beginner, Intermediate, etc. 

 
image

To test that they are displaying, you can turn on the UDF fields in the board edit page. 

I did find one UDF type that needs to be updated. It looks like UDFs with date fields are now being sent as datetimes. Example: 
image

kti-daniel:
Thanks Cameron, 

As I investigated, we have 3 UDFs currently available, which are ones for ticket, task, and project. All these have type information that could be datetime. I guess it seems delivered as date string before, but not for now (just guessing). I am not familiar to retrieve using SOAP.

In our dev env., only ticket_udfs have datetime format. Therefore, I might guess with carefulness customers using UDF with datetime type are looking at the value as it was delivered without any modification. (i.e. They can see simple date format only if SOAP does.) It is totally depends on SOAP API response.

So, I can guess adding this code snippet is allowable without any side-effect. Please review them again, and confirm this. Thanks again.

            if udf_field['type'] == 'datetime':
                udf_value = udf_value[:10]

djautotask/sync_rest.py
```diff
@@ -296,7 +340,7 @@ def get_page(self, next_url=None, *args, **kwargs):
 
 
 # ParentSynchronizer: using SOAP API
-class TicketSynchronizer(SyncRecordUDFMixin, Synchronizer, ParentSynchronizer):
+class TicketSynchronizer(SyncRestRecordUDFMixin, Synchronizer, ParentSynchronizer):
```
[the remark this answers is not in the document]
kti-cameron: This didn't get changed but I'll handle it.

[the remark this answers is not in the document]
kti-daniel: Oh, thanks. I forgot.

kti-cameron:
I can confirm that SOAP responses are returning dates not datetimes. In future if you need to query the SOAP API, you can do so using the atws library. Either take a look at the sync code or docs to see how to write a query https://atws.readthedocs.io/usage.html  .

So your fix should be good. I'll merge shortly.

PR #127, opened 2021-05-21
Change task sync request to POST

kti-cameron:


djautotask/api_rest.py
```diff
@@ -291,21 +291,33 @@ def fetch_resource(self, next_url=None, retry_counter=None, *args,
...
-                logger.debug('Making GET request to {}'.format(endpoint))
-                response = requests.get(
-                    endpoint,
+                logger_message = 'Making {} request to {}'.format(
+                    request_method.upper(), endpoint)
+                if body:
+                    logger_message = \
+                        '{}. Request body: {}'.format(logger_message, body)
+
+                logger.debug(logger_message)
```
kti-daniel, line 308: 1. if body:
    Do we need body here? I think in the case of POST, we can add log to show the contents of POST body. In this case, the code can be simplified in the future:

```
        self.build_query_string(**kwargs)

        return self.fetch_resource(next_url, method='post',
                                   body=self.QUERYSTR, *args, **kwargs)
```
to
```
        return self.fetch_resource(next_url, method='post', *args, **kwargs)
```
2. logger_message in line 305 override line 302. Do you intend to concatenate logs here ?

kti-cameron: 1. My intent was to have a debug log to show the POST conditions since when using GET you can see the conditions in the URL. If you can improve the code and still ensure we can see debug logs showing query conditions for both GET and POST let me know.

I don't understand how we can simplify:
```python
        self.build_query_string(**kwargs)

        return self.fetch_resource(next_url, method='post',
                                   body=self.QUERYSTR, *args, **kwargs)
```
to 
```python
        return self.fetch_resource(next_url, method='post', *args, **kwargs)
```
can you explain that a bit more?

kti-cameron: 2. Yes I want to concat those logs. Maybe I'm missing something - I thought line 305 would take the existing logger_message and append the request body conditions/additional text.

kti-daniel:
```
self.build_query_string(**kwargs)
body=self.QUERYSTR
```
This can be done inside of `self.fetch_resource` when self.QUERYSTR is None. So, we can just say:
`self.fetch_resource(next_url, method='post', *args, **kwargs)`.

I thought you need to pass body parameter to display POST body for logging. If this is the only purpose, I think we can just say:

`if body:` to `if request_method.upper() == 'POST':` or something like this.

Do I miss something important? Then, please let me know.

kti-cameron: Thanks for clarifying. I think I get it now and yeah that does simplify things. I'll work on making those changes.

kti-cameron: Nice suggestions. I think it's cleaner now.

kti-cameron: Just added another fix cause I realized I was send QUERYSTR as the body even on GET requests. Should be fixed now.

kti-daniel:
If you don't mind, can I made some small change?

In the case of https://webservices2.autotask.net/atservicesrest/v1.0/Tickets/7682, the last parameter can be None at `log_message(self, endpoint, method, body)`

So, I will add this line in the log_message method.

`        body = body if body else '' `

kti-cameron:
Sure that's fine :+1:

kti-daniel:
All looks good. Not testing. After that, it will be merged soon.

PR #128, opened 2021-05-21
Uncaught exceptions when making requests with AT REST API

kti-daniel:


kti-cameron:
Looks like these changes will be added in https://github.com/KerkhoffTechnologies/django-autotask/pull/129 . I'll just close this PR

kti-daniel:
Oh, Oh~ 

It looks similar, but it is not.

```
        'assigned_resource': 'assignedResourceID',
        'assigned_resource_role': 'assignedResourceRoleID',
```
To properly handling dragging exception, the codes above are necessary, which are not included in #128 . I was absent for a while between 5 to 6, so I cannot respond your comment at the proper time. Sorry.

I will patch the code snippet above to #128 . After that, I feel we need to update all clients. 
Have a good long weekend.

kti-cameron:
So sorry about that. I don't know how I missed those 2 assigned_resource lines. These changes are only in our training instances but I'll make sure #1802 is merged in promptly.

kti-daniel:
Thumbs up! Thanks Cameron.

PR #134, opened 2021-06-09
MOD: Solve query limitation in OR query

kti-daniel:


djautotask/sync_rest.py
```diff
@@ -140,6 +146,45 @@ def get_single(self, *args, **kwargs):
     def _assign_field_data(self, instance, api_instance):
         raise NotImplementedError
 
+    # another approach for sync to resolve an issue of AT REST API
+    # query limitation(up to 500), especially when OR & IN conditions are used.
+    # Page calls & DB update should be done here while conditions are hold
+    # because IN-condition clauses are divided, used, and changed during sync.
+    def fetch_records_by_altered_condition(self, next_url, results,
```
kti-cameron, line 153: I've only taken a pretty quick look at this but from what I can tell it does work but I find it difficult to read.

Could we instead add something along of the lines of:
```python
    def fetch_records(self, results):
        project_ids = list(self.get_active_ids())

        next_url = None
        while project_ids:
            batch_size = self.changing_condition_num
            batch = project_ids[:batch_size]

            self.api_conditions += ['projectId', batch, 'in']
            del project_ids[:batch_size]

            logger.info(
                'Fetching {} records'.format(
                    self.model_class.__bases__[0].__name__)
            )
            api_return = self.get_page(next_url, next_url,
                                       conditions=self.api_conditions)
            page = api_return.get("items")
            next_url = api_return.get("pageDetails").get("nextPageUrl")
            self.persist_page(page, results)

            if not next_url:
                break

        return results
```
I didn't test the code above but I think it would do a similar job with a few tweaks. It uses a similar pattern to  the `BatchQueryMixin` that is used for the SOAP API. 

Let me know if I'm missing the reason why this approach wouldn't work though.

djautotask/sync_rest.py
```diff
@@ -24,6 +24,12 @@ class Synchronizer:
     lookup_key = 'id'
     last_updated_field = 'lastActivityDate'
 
+    # optional member variables for fetch_records_by_altered_condition method
+    changing_condition_pool = None
+    changing_condition_field_name = None
+    get_page_method_name = None
+    changing_condition_num = 100
```
kti-cameron, line 31: Does this determine the size of the batches? If so, there is already a Constance option for controlling batch size `DJAUTOTASK_BATCH_QUERY_SIZE` and we might be able to use that. If it doesn't conflict with anything then we should use it.

djautotask/sync_rest.py
kti-daniel: Hi Cameron, I took a quick look at `BatchQueryMixin`, and it looks doing similar job. But, I am not sure if I understand in detail how `BatchQueryMixin` works.

The main purpose of my work was two things:
1) to divide IN-clause query list and ask API request
2) count the divided-asked query result

Can I ask if these two purposes can be achieved from `BatchQueryMixin`?

kti-daniel: In graphic,

[As is]

kti-cameron: `BatchQueryMixin` doesn't do exactly what you describe but the result is the same - break up a query with a lot of IDs into smaller batches. But maybe it's not a great example because the queries for SOAP are very different than what we have to do for REST. The main thing I wanted to point out was pulling out batches of IDs from a list then adding them to the request conditions.

There's nothing strictly wrong with your method, I just find to be more complex than I think it needs to be. I'd prefer if we could come up something a bit simpler.

Since we want this resolved quickly, I did a bit of work on it myself. Here's an updated snippet.

Can you review/give it a try? The only bit that is not working right now is in `_fetch_resource`. It needs to be changed somehow because at the moment it only builds self.QUERYSTR once and we need it to update for the next batch of IDs.

```python
    def fetch_records(self, results):
        project_ids = list(self.get_active_ids())

        next_url = None
        while project_ids:
            batch = project_ids[:self.batch_size]

            # If projectId is already in the condition list, remove
            # so we can add the next batch of projectIds
            if 'projectId' in self.api_conditions[-1]:
                self.api_conditions.pop(-1)

            self.api_conditions += [
                ['projectId', batch, 'in']
            ]
            del project_ids[:self.batch_size]

            logger.info(
                'Fetching {} records'.format(
                    self.model_class.__bases__[0].__name__)
            )
            api_return = self.get_page(next_url, next_url,
                                       conditions=self.api_conditions)
            page = api_return.get("items")
            next_url = api_return.get("pageDetails").get("nextPageUrl")
            self.persist_page(page, results)

            if not project_ids and not next_url:
                break

        return results
```

kti-cameron: Note: 
To get that snippet to work you'll need to add:

```python
settings = DjautotaskSettings().get_settings()
self.batch_size = settings.get('batch_query_size')
```
and remove:
```python
self.api_conditions += [
            ['projectId', list(self.get_active_ids()), 'in']
        ]
```
in `TaskSynchronizer` `__init__`

kti-daniel: Hmm.. it could be solved with this bit `self.client.QUERYSTR = None`, which already exists in `fetch_records_by_altered_condition`. 

But, I have one concern here. If we override only `fetch_records` for TaskSynchronizer as the codes shows above, we would do the same pattern of things for similar cases. In ProjectSynchronizer also has IN-clause, so the API has a potential probability to override their fetch record alike, which produces long similar codes in our sync_rest.py. 

So, the general part of the codes in fetch record is extracted as `fetch_records_by_altered_condition`. In this case, the fetch record in each Synchronizer, small part of codes are required.

    def fetch_records(self, results):
        next_url = None
        return self.fetch_records_by_altered_condition(next_url, results)


However, I feel that this approach is not the best from the point of OO, when you introduce `BatchQueryMixin`. We don't have the same pattern in CW REST, so this work would be the first approach for dividing query in REST API. 

If we divide calls inside get method as current SOAP does, we can use fetch_records (DB update and count results) & get_page (subsequent page call of each query) as it is. So, accepting & following your introduction of `BatchQueryMixin`, dividing query and making request can be moved from `fetch_records_by_altered_condition` to get method.

I will share the work after your lunch.

kti-daniel:
The same name of `BatchQueryMixin` has been added to play a role of making divided API requests. It overrides `get` method, so we can take the same name of `get`. In addition, the name of `BatchQueryMixin` is familiar to us, so I decided to use the same name.

For this part,

            for idx, c in enumerate(self.api_conditions):
                if isinstance(c[0], str) and c[0] == self.condition_field_name:
                    self.api_conditions[idx][1] = batch_condition

in REST API, they have a JSON structure, and the relevant API condition can be placed any position of JSON structure. So, I think it is still required. But, the outer query is changed following while & del pattern.

Now you can have another look. Thanks for your review!!!

kti-daniel:
FYI. there are variables named `batch_size` and `batch_query_size`, so to avoid confusion, the exact name `'batch_query_size'` is used.

    def get_settings(self):
        request_settings = {
            'timeout': 30.0,
            'batch_size': 50,
            'max_attempts': 3,
            'keep_completed_hours': 8,
            'batch_query_size': 400,
        }

PR #136, opened 2021-06-28
Handle task update with null project

kti-cameron:


djautotask/models.py
```diff
@@ -618,6 +618,12 @@ def __str__(self):
         return self.title
 
     def update_at(self, **kwargs):
+
+        if not self.project:
+            raise api_rest.AutotaskAPIClientError(
```
kti-sam, line 623: As nice as it would be to just catch this exception, `AutotaskAPIClientError` is definitely not the exception to use. As this isn't a problem with the API client. 

Is a task being in our DB a result of AT allowing tasks without projects or is that because we don't cascade delete tasks without a project? If the cause of a projectless task is somehow our fault its probably better to handle it entirely in our app rather than django-autotask.

kti-cameron: Yeah you're right. I was thinking just 'client' not APIClient for some reason. So client makes request to API but it can't build the URL so that's the client's fault - raise the exception. 

I think it's our fault that we allow a task without a project in our DB. Docs say project is required: https://www.autotask.net/help/DeveloperHelp/Content/APIs/REST/Entities/TasksEntity.htm . I'll test creating/updating a task via the API to see if the it enforces having a project associated with the task. If the API does have validation for this, I figure I'll just remove the `nul=True` from the `project` FK on Task and cascade delete. 

I think that should solve the issue. Let me know if are aware of reasons why this wouldn't be a good approach.

kti-cameron: Confirmed. You have to provide a projectId to create a task via the API or else the POST will fail. 

So I think we can remove the null=True from the project FK on Task and add cascade delete. 

That would mean we handle the problem before it becomes a problem in django-autotask by simply not saving tasks if they somehow don't have a project. If you don't see a problem with that I'll make that change.

kti-sam: Due to a bug in the tracker with this version of django (known issue) we may not be able to cascade delete without getting a database error. So, be aware of that.

kti-cameron: Needed a data migration so added that. It just removes any tasks from the DB if they have a null project. Tested syncing a task without a project (should be impossible) but checked anyway - we'll just catch the integrity error and keep syncing. When a project is removed during a full sync I didn't see any problems with tasks not getting cascade deleted. 

Ready for another look.

djautotask/migrations/0092_auto_20210629_1124.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def process_tasks(apps, schema_editor):
+    Task = apps.get_model('djautotask', 'Task')
+
+    # Remove tasks without a project in preparation for migration
+    # removing null=True from project foreign key on Task model.
+    null_project_tasks = Task.objects.filter(project__isnull=True)
+    for task in null_project_tasks:
+        task.delete()
```
kti-sam, line 13: This actually worries me. In theory its totally fine, but we have that efficiency problem when trying to delete many objects with ranks attached. It's possible this could cause a deadlock during migration or make it run for an hour in a bad case. I'll make up a dummy db with data to mimic that and test it before I merge.

kti-matt: Maybe first check if we actually have any customers with a large number of tasks. We may not.

kti-sam: Ah, good idea!

kti-cameron: Yeah good point and thanks for checking that. I was banking on it being quite rare to have a large number of tasks with a null project FK.

kti-sam: No clients with Tasks and null projects, should be all good! I'll finish up review and merge shortly.

kti-sam:
Getting unit test fails

kti-cameron:
Yeah forgot that tests would be affected by that change. Ugh doesn't look like an easy fix either. I'm seeing some weird test fails. I'll dig into it and update when done.

kti-cameron:
Fixed now and ready for a look. It wasn't as bad as I thought. The main problem was `test_full_sync` and a mock not getting cleaned up due to the failure. That caused a whole lot of unrelated looking test fails :facepalm:

djautotask/models.py
kti-sam: I don't think you pushed your changes

kti-sam:
I don't think you pushed your changes

kti-cameron:
Sorry. Pushed to wrong branch. Should be updated now.

PR #141, opened 2021-07-23
ADD: contact remote loading

kti-daniel:


djautotask/__init__.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-VERSION = (0, 0, 127, 'alpha')
+VERSION = (0, 0, 129, 'alpha')
```
kti-cameron, line 2: Just needs to be bumped to 130

PR #148, opened 2021-08-17
MOD: complex list to class to improve logic

kti-daniel:


djautotask/api_rest.py
```diff
@@ -118,6 +118,40 @@ def get_zone_info(username):
...
+        '<': 'lt',
+        '<=': 'lte',
+        '^=': 'beginsWith',
+        '=$': 'endsWith',
+        'contains': 'contains',
+        'exists': 'exist',
+        'not exists ': 'notExist',
+        'in': 'in',
+        'not in': 'notIn',
+    }
```
kti-sam, line 138: What is the benefit of translating the AT operators to a custom operator list?

djautotask/api_rest.py
```diff
@@ -223,32 +266,42 @@ def get_headers(self):
...
+            conditions = kwargs['conditions']
+
+            if hasattr(conditions, 'operands'):
+                filter_obj = {
+                    "op": conditions.operator,
+                    "items": [
+                        self._build_filter_unit(c)
+                        for c in conditions.operands
+                    ]
+                }
```
kti-sam, line 284: Since you designed a specific class to handle API conditions, the logic for building those conditions should be put in that class as well. 

The way it is currently is so highly coupled its hard to tell the reason for the separate class at all. As the `AutotaskAPIClient` class has to check the `ApiCondition` instance if it has a specific attribute, and then builds the condition itself. There are even helper methods in `AutotaskAPIClient` for building filters rather than in `ApiCondition`.

djautotask/sync_rest.py
```diff
@@ -462,17 +480,13 @@ def __init__(self, *args, **kwargs):
...
-                    ]
-                }
-            ]
-        ]
+        self.api_conditions = ApiCondition(
+            op='or',
+            operands=[
+                ApiCondition(self.completed_date_field, self.completed_date,
+                             op='>'),
+                ApiCondition('status', models.Status.COMPLETE_ID, '!=')
```
kti-daniel, line 488: Oh,  there is no benefit to use general operator here. The missing `op=` (i.e. `op='!='`)  makes deletion of tickets except those with status `complete`. I will role back this one as you said. It is prone to make a big mistake.

PR #151, opened 2021-09-09
2131 convert at service call relevant sync and api classes to rest

kti-daniel:


djautotask/sync_rest.py
```diff
@@ -781,6 +790,225 @@ def _assign_field_data(self, instance, object_data):
...
+        Make a request to Autotask to create a ServiceCall.
+        """
+
+        body = {
+            'companyID': kwargs['account'].id,
+            'companyLocationID': kwargs['location'].id,
+            'startDateTime': kwargs['start_date_time'],
+            'endDateTime': kwargs['end_date_time'],
+            'status': kwargs['status'].id,
+            'description': kwargs['description'],
```
kti-sam, line 849: We now have a pattern for mapping the model fields to the API field names via the `EDITABLE_FIELDS` dicts on the models. Since synchronizers already have a reference to their `model_class` it wont be difficult to merge that behaviour  so we don't have to duplicate figuring out the field names for creating AND updating.

djautotask/utils.py
```diff
@@ -2,17 +2,26 @@
...
+        # service_call_task shows errors w/ the value of batch_query_size
+        # larger than 215. The error: 404 - File or directory not found.</h2>
+        # <h3>The resource you are looking for might have been removed,
+        # had its name changed, or is temporarily unavailable.
         if hasattr(settings, 'DJAUTOTASK_CONF_CALLABLE'):
             request_settings.update(settings.DJAUTOTASK_CONF_CALLABLE())
+            request_settings['batch_query_size'] = min(
+                self.force_batch_query_size,
+                request_settings.get('batch_query_size')
+            )
```
kti-sam, line 25: I don't think any of this should be in `DjautotaskSettings`. If I am reading this correctly it will limit ALL batch queries to 215, even the ones that would work with a higher number.

Instead you should just have `ServiceCallTaskSynchronizer` override its __init__ method, and then you can just set `self.batch_query_size` to be 215 there with this `min` method.

djautotask/api_rest.py
```diff
@@ -327,37 +327,49 @@ def get_headers(self):
...
-                    )
+            body = self._format_request_body(body, key, value)
 
-                else:
-                    body.update(
-                        {key: str(value) if value else ''}
-                    )
+        return body
+
+    def _format_inserted_fields(self, inserted_fields):
```
kti-sam, line 343: Editing and creating should be functionally the same operation right? Do we need this division of methods between updating or creating?

djautotask/sync_rest.py
```diff
@@ -781,6 +790,225 @@ def _assign_field_data(self, instance, object_data):
...
+        }
+        instance = super().create(body)
+        created_item = self.get_single(instance['id'])
```
kti-sam, line 852: Isn't `instance` already the created record? Do you need to fetch it again with `get_single`? Can't you just pass it into `update_or_create_instance`?

djautotask/utils.py
kti-daniel: That's a good point. Will fix that.

djautotask/sync_rest.py
kti-daniel: There are differences between them. Most of all, EDITABLE_FIELDS are restricted, it is not in the case of creation. But, as you said, duplications can be improved a little more at the level of model. Will fix that.

kti-sam: Ok, that begs the question then. Why is editable fields restricted? Does it need to be? Who are we restricting?

If the answers to those questions are "ourselves", lets change `EDITABLE_FIELDS` to something like `AUTOTASK_FIELDS` and then add everything we intend to edit or create. How does that sound?

kti-daniel: Yeap, I strongly agree. :+1:

kti-daniel: @camsweeney or @kti-cameron, can I ask your idea here? We have for now `EDITABLE_FIELDS` and it decides which fields are possible to be edited. Instead, in the case of introducing `AUTOTASK_FIELDS`, we will use this dictionary just for mapping. (i.e. any fields can be asked for edit if this mappings are working) 

Do you have any recommendation or idea here?

kti-sam: @kti-cameron To give more context. The main question is: Is there a purpose for `EDITABLE_FIELDS`? Why do we have this safety net, and can we just get rid of it in favour of a dictionary that just maps db column names to the API field names?

kti-cameron: The description for the issue where `EDITABLE_FIELDS` was added is at https://git.ops.topleft.team/apps/django-topleft/-/issues/1797 .

So Matt suggested adding something like EDITABLE _FIELDS. I'm not sure exactly what Matt's rationale was for keeping a list of fields we can edit vs fields we can't. But I'd guess it has something to do with avoiding inadvertent updates to AT. 

For my part, I don't think it's a bad idea for us to have to be very specific about what TopLeft is allowed to update in AT.

kti-sam: Ok, that makes sense to me. So that isn't really a safety net at all. It's just there to be a field mapping to translate db columns to API field names in an easy way. Its not there to be specific about what TopLeft can update (and even if it was, this would we the wrong place for it as we don't want limitations in TopLeft to have ANY effect on `django-connectwise` or `django-autotask`.)

Ok, sounds good. Lets change `EDITABLE_FIELDS` to `AUTOTASK_FIELDS`

kti-cameron: No, it is a safety net. Fields not it `EDITABLE_FIELDS` won't be added to the object to update in AT. See https://github.com/KerkhoffTechnologies/django-autotask/blob/f9cdbd6a76cca1b84a71175ad95c1d2c801d7436/djautotask/api_rest.py#L337

But I take your point about django-autotask not being the best place to keep these restrictions.

kti-sam: Ok, then I say I 100% disagree with the need for this safety net. If we already have the list of changed fields, if that list is wrong due to some error in TopLeft, then it should be handled in TopLeft. 

BUT this still begs the question why is it there, Is there some piece of the puzzle missing here? Are we expecting the data in `changed_fields.items()` to be untrustworthy? I don't see why we can't take `changed_fields` at face value...

kti-daniel: I can say that the current safety net has a role to force TopLeft app (using djautask) to pass only edit-allowable field values.

But I can take the point that we can trust `changed_fields`, which is developed by us.

kti-sam: No, I am saying its absolutely not ok for django-autotask to be a safety net for TopLeft. We should not develop django-autotask to accomodate TopLeft, we should develop TopLeft to accomodate django-autotask.

Also, I am not making the point `changed_fields` is something we can trust, I am saying there might be a reason we can NOT trust `changed_fields` that I am unaware of, and I am currently looking for that reason.

kti-sam: Ok, while I can not find a reason we can't trust changed fields, or a precise reason for the safety net, I don't think editable fields is intended to segregate the logic between creating and editing, I think that was just a name picked at the time. So I think we can leave the EDITABLE_FIELDS name in, and even the `if field in api_entity.EDITABLE_FIELDS: ` line in, but @kti-daniel , you can also add in all the fields we need to use to create a model as well :+1:

kti-daniel: Okay, I will do. Thanks for the discussion, Sam and Cameron.

djautotask/sync_rest.py
kti-daniel: Applied. You can now have another look. Thanks~

djautotask/api_rest.py
```diff
@@ -327,37 +327,50 @@ def get_headers(self):
...
-                else:
-                    body.update(
-                        {key: str(value) if value else ''}
-                    )
+    def _format_inserted_fields(self, api_entity, inserted_fields):
+        body = dict()
+
+        for field, value in inserted_fields.items():
+            if field in api_entity.AUTOTASK_FIELDS:
+                key = api_entity.AUTOTASK_FIELDS[field]
```
kti-sam, line 347: Why do you have `AUTOTASK_FIELDS` and `EDITABLE_FIELDS` now? Just have the one, `EDITABLE_FIELDS` and put everything we need to edit or update in there. Then merge `_format_inserted_fields` and `_format_changed_fields` into _format_fields like we discussed on zoom. Like that example I showed you where you can merge the methods by changing the body initialization to:

`body = { 'id': api_entity.id } if api_entity else dict()`

kti-sam: But that's what I spent time researching yesterday. To make sure it would be ok, and it looks like it is. If you can find a concrete reason we can't just merge them into one, then ok. But if you can't find anything please add the fields to `EDITABLE_FIELDS`. I'm fine with you making changes, but you have to have a reason for it, you can't just guess there might be a problem, you need evidence.

kti-daniel: Hmm.. then, it means editable fields doesn't mean "edit-allowable" fields, and it doesn't have a role of confining "edit-allowable fields". Then, my another choice could be renaming the value to `AUTOTASK_FIELDS` instead. 

Can I ask this to @kti-cameron (or @camsweeney ) again? :smiley:

kti-sam: That's fine too, I just figured by keeping it EDITABLE fields it would be nicer to keep pairity with django-connectwise. I don't care what you call it as long as its merged.

kti-daniel: Thanks Sam, in AT case, the concept of `EDITABLE_FIELDS` is extended, so `AUTOTASK_FIELDS` could be better. The combining is applied, and ready again.

djautotask/sync_rest.py
```diff
@@ -781,6 +804,203 @@ def _assign_field_data(self, instance, object_data):
...
+        instance.id = object_data['id']
+        self.set_relations(instance, object_data)
+
+        return instance
+
+    def create(self, **kwargs):
+        kwargs.update({
+            'parent': kwargs.get('service_call')
+        })
+        return super().create(**kwargs)
```
kti-sam, line 897: parent should be passed in as an argument like on the update method, we shouldn't need to update `kwargs`. Its part of the childapimixin so we always know a parent will exist.

djautotask/api_rest.py
```diff
@@ -534,9 +541,16 @@ def get_child_url(self, parent_id):
...
-        body = self._format_request_body(instance, changed_fields)
+        body = self._format_fields(instance, changed_fields)
         return self.request('patch', endpoint_url, body)
 
+    def create(self, instance, **kwargs):
+        parent = kwargs.pop('parent')
+        endpoint_url = self.get_child_url(parent.id)
+        body = self._format_fields(instance, kwargs)
+        response = self.request('post', endpoint_url, body)
+        return response.get('itemId'), body
```
kti-sam, line 552: It took me a while to figure out why you are passing the body back as the response, as that is very bad. We could get some very major errors and bugs that way. 

When I said you should pass the response into `update_or_create_instance`, I meant pass the response we get from Autotask. I thought both the [service call tickets child](https://webservices2.autotask.net/atservicesrest/swagger/ui/index#/ServiceCallTicketsChild/ServiceCallTicketsChild_CreateEntity) and the [service call tickets]() endpoints return the created record in the response, but it looks like only the main [ServiceCallTickets](https://webservices2.autotask.net/atservicesrest/swagger/ui/index#/ServiceCallTickets/ServiceCallTickets_Query) enpoint has that in the response, because its a query not a create operation. That's really annoying on their part.

If this worked like connectwise, a successfuly POST operation returns the record we just created, which would make our life very easy. Why does AT have to be dumb...

I think the safest way to handle this is to undo the changes you made from this comment: https://github.com/KerkhoffTechnologies/django-autotask/pull/151#discussion_r706500844 , while it isn't the most efficient request wise, it is the safest way to ensure we keep synchronized.

kti-daniel: Yeap, I thought you already had checked that response and asked that. Applied and review again.


cf. One more thing for our KB: 

> but it looks like only the main ServiceCallTickets enpoint has that in the response, because its a query not a create operation.

This post doc isn't relevant for ServiceCallTickets creation. See [this](https://webservices2.autotask.net/atservicesrest/swagger/ui/index#/ServiceCallTicketsChild/ServiceCallTicketsChild_CreateEntity) or [that one](https://git.ops.topleft.team/apps/django-topleft/-/issues/2131#note_32352), too.

djautotask/api_rest.py
```diff
@@ -521,7 +521,7 @@ def update(self, instance, changed_fields):
 
     def create(self, instance, **kwargs):
         body = self._format_fields(instance, kwargs)
-        # API returns the newly created id
+        # API returns only newly created id
```
kti-sam, line 524: Also a good distinction to make in the comments, good one.

PR #152, opened 2021-09-24
MOD: TypeError in fetch_sync_by_id

kti-daniel:


djautotask/sync_rest.py
```diff
@@ -181,6 +181,9 @@ def persist_page(self, records, results):
                     results.updated_count += 1
                 else:
                     results.skipped_count += 1
+            except TypeError as e:
+                logger.warning('{}'.format(e))
+                continue
```
kti-sam, line 186: This can be removed, it is impossible for this to happen. This is running for each element in `records` so we know it won't be `None`. The problem only exists in `fetch_sync_by_id`.

djautotask/sync_rest.py
```diff
@@ -206,8 +209,13 @@ def _set_datetime_attribute(self, instance, attribute_name):
...
     def fetch_sync_by_id(self, instance_id):
         api_instance = self.get_single(instance_id)
-        instance, created = \
-            self.update_or_create_instance(api_instance['item'])
+        try:
+            instance, created = \
+                self.update_or_create_instance(api_instance['item'])
+        except TypeError as e:
+            logger.warning('{}'.format(e))
+            instance = None
```
kti-sam, line 217: Instead of trying to create something we know will fail (because `get_single` returned a null instance) lets check if it exists and raise a `AutotaskRecordNotFoundError` if it doesn't. That way calling code can pick that up and handle it as it chooses. Otherwise we have to adapt all current code to detect when `fetch_sync_by_id` returns `None`, which is already a big change as no code expects that to ever return `None`. 

Then just make sure everything that calls `fetch_sync_by_id` catches `AutotaskRecordNotFoundError`. I'm pretty sure its just callbacks and maybe one or two other things. Not a lot.

To be fair this is Autotasks fault. If we request a record and it doesn't exist, it should be kind enough to give us a 404. Yeesh.

djautotask/sync_rest.py
```diff
@@ -610,7 +618,7 @@ def _assign_field_data(self, instance, json_data):
 
     def fetch_sync_by_id(self, instance_id):
         instance = super().fetch_sync_by_id(instance_id)
-        if not instance.status or \
+        if instance and not instance.status or \
```
kti-sam, line 621: I guess just revert this after you make the changes for the other comment about 404's

djautotask/sync_rest.py
kti-daniel: I tested the following code snippet. Yes, Sentry doesn't complain in this code, but when AT-side is unstable (for example, unexpected errors such as seen at `fetch_sync_by_id` or max-request occurrences, and so on), it could be some kind of safety net.

image

But, I can agree with you because we haven't seen any errors yet here.

djautotask/sync_rest.py
kti-daniel: Regarding `api_instance`, it should return normally as follows (from the responses of postman):

```
{
    "items": [],
    "pageDetails": {
        ...
    }
}
```

In other words, in usual cases, api_instance['item'] should be empty list, and it can be iterable. 

But, to be surprise, it seems return as follows occasionally:

```
{
    "items": null,
    "pageDetails": {
        ...
    }
}
```

Furthermore, regarding return of `fetch_sync_by_id`:
In the case that get_single returns instance, the code return instance. So, if it doesn't exist, it can return None logically.
Luckily, their return is only used in one line (line 620) for now.

image

So, I hope to hear your opinion once again. 

However, I have got some idea from your comments, so I fixed the code a little bit more and pushed that, too.

djautotask/sync_rest.py
kti-sam: Also I don't think the AT-side being unstable is guaranteed to be a TypeError, is it? We shouldn't just guess at exceptions.

djautotask/sync_rest.py
kti-sam: Ok, this is worth talking about. Can you provide more information? Under what conditions does this occational case happen? Can you reproduce it reliably? These are important things to know before deciding what to do.

kti-sam: Does it only return null if we use `get_single/fetch_sync_by_id`? If so then my previous point about `AutotaskRecordNotFoundError` still stands.

djautotask/sync_rest.py
kti-daniel: Okay, I will remove that.

PR #160, opened 2021-11-04
MOD: Fix Autotask API never clears web url cache

kti-daniel:


djautotask/api.py
```diff
@@ -22,6 +22,10 @@
 NO_RECORD_ERROR_MESSAGE = \
     'No message. No matching records found. The logged in Resource may not ' \
     'have the required permissions to delete the data.'
+CACHE_KEYS = {
+    'url': 'zone_info_url',
+    'webUrl': 'zone_info_webUrl'
```
kti-sam, line 27: What's the benefit of putting these into a dict to access them as different names? On line 425 you go back to just using 'zone_info_url' again anyway. If there isn't a benefit to this you can just remove it and use 'zone_info_url' and 'zone_info_webUrl'.

kti-daniel: Okay, just a min. The AT response is as follows:

image

So, it could be better to make consistency using `url` and `webUrl`

kti-daniel: Fix applied, and review is ready again

djautotask/api.py
```diff
@@ -110,6 +113,8 @@ def get_zone_info(username):
         response = requests.get(endpoint_url)
         if 200 == response.status_code:
             resp_json = response.json()
+            # Whenever get_zone_info is triggered, update cache
+            update_cache(resp_json)
```
kti-sam, line 117: This is not what we want to do here, get _zone_info is triggered every time we initialize a synchronizer, which is almost every time we make a request to the AT API now. There isn't a point in keeping a cache if its refreshed nearly every time.

djautotask/api.py
```diff
@@ -76,23 +76,26 @@ def get_cached_url(cache_key):
...
-    return _get_connection_url('url')
+def update_cache(json_obj):
+    cache.set('url', json_obj['url'])
+    cache.set('webUrl', json_obj['webUrl'])
 
 
-def get_web_connection_url():
-    return _get_connection_url('webUrl')
+def get_api_connection_url(force_fetch=False):
+    return _get_connection_url('url', force_fetch)
```
kti-sam, line 85: `url` is actually a poor cache key, and we shouldn't have used that. The django cache is shared so something else may try to use the `url` key accidentally in the future. Can you make this more specific? Like 'zoneUrl'?

djautotask/api.py
```diff
@@ -76,23 +76,26 @@ def get_cached_url(cache_key):
...
-    if not api_url_from_cache:
+
+def _get_connection_url(field, force_fetch=False):
+    api_url_from_cache = get_cached_url(field)
+
+    if not api_url_from_cache or force_fetch:
         try:
             json_obj = get_zone_info(settings.AUTOTASK_CREDENTIALS['username'])
             url = json_obj[field]
-            cache.set(cache_key, url, timeout=None)
```
kti-sam, line 95: This is actually the timeout mentioned in the issue description "Done when" section. It shouldn't be removed, it should be updated to some finite amount of time. Let's say 10 minutes.

djautotask/api.py
kti-daniel: Oh, really? Maybe there seems to be some confusion here.

For now, the `get_zone_info` has been used as a last call to refresh cache info.

    if not api_url_from_cache or force_fetch:
        try:
            json_obj = get_zone_info(settings.AUTOTASK_CREDENTIALS['username'])
            url = json_obj[field]
        except AutotaskAPIError as e:
            raise AutotaskAPIError('Failed to get zone info: {}'.format(e))

kti-daniel: Maybe you are confused w/ `get_api_connection_url` and `get_zone_info`

image

Please let me know if I misunderstand something. :)

kti-sam: Yes, I think you are right. I believe I mixed up `get_api_connection_url` and `get_zone_info`. Sorry about that. Just the other two points about changing the cache keys and increasing the timeout to 10 minutes.

kti-daniel: Thanks~ It has been fixed and review again.

djautotask/api.py
```diff
@@ -73,26 +73,31 @@ def retry_if_api_error(exception):
 
 
 def get_cached_url(cache_key):
-    return cache.get(cache_key)
+    return cache.get('Zone{}'.format(cache_key.capitalize()))
```
kti-sam, line 76: Why the call to capitalize? Why not just be consistent when you use the key in the code? Its a small change though, I'll just make it and merge the issue.

djautotask/tests/mocks.py
```diff
@@ -17,7 +17,7 @@ def create_mock_call(method_name, return_value, side_effect=None):
 
 def init_api_rest_connection(return_value=None):
     method_name = 'djautotask.api.get_api_connection_url'
```
kti-sam, line 19: I found a problem in this unit test you added for the REST switch while making my changes to the issue. You are testing a method too high in the call stack now. By testing here instead of zone info, you are just setting the cache rather than testing the logic of whether it should fetch from the cache or refresh the data from the API. I'll update this too.

kti-sam:
Ok @kti-daniel , I've made some changes. You can go over my changes to review and see what I meant about testing the right method. You can merge this issue when you are satisfied.

kti-daniel:
Yes, it looks good. And I just added for a small improvement as follows and will merge it into:

                if request_retry_counter['count'] <= self.MAX_401_ATTEMPTS:
                    cached_url = get_cached_url(AT_URL_KEY)

PR #163, opened 2021-12-28
ADD: Sync AT task category picklist and set on tasks

kti-daniel:


djautotask/models.py
```diff
@@ -647,6 +652,9 @@ class Task(ATUpdateMixin, TimeStampedModel):
     department = models.ForeignKey(
         'Department', blank=True, null=True, on_delete=models.SET_NULL
     )
+    task_category = models.ForeignKey(
```
kti-sam, line 655: Change this to just `category`. We already know its a task category, and if it has the same field name as ticket categories that will help with generalization.

PR #167, opened 2022-01-14
use license_type class to determine impersonation possibility

kti-daniel:


djautotask/api.py
```diff
@@ -588,6 +588,19 @@ def delete(self, instance, parent=None):
         # AT sends deleted_id or 500 in the case failure instead of 404 or 204
         return response.get('itemId')
 
+    def _set_impersonation_id(self, impersonation_id, target=None):
+        from djautotask.models import Resource
+        checked_impersonation_id = None
+        if impersonation_id:
+            try:
+                resource = Resource.objects.get(id=impersonation_id)
```
kti-sam, line 596: I said you should just pass the resource instance directly to the update method. You already have it everywhere the update method is called, why just pass in the resource ID if you are just going to query for it again?

Can you explain how a mid-file import in a function to query for an instance you could have passed in is better than just passing the instance in?

kti-daniel: I thought `**key ` can be still used without change.

As you said, it sends additional query. I will fix that, too.

kti-sam: Ok, I explained that I have a problem with that whole `**key` method of passing in just the ID already. It's better to just pass in the resource in cases where applicable. That way there are two cases from TopLeft's perspective: 
* Impersonation is requested -> pass resource as an argument
* Impersonation is NOT requested -> don't pass the resource

Then all the logic if it CAN use impersonation exists in `django-autotask`, rather than TopLeft having all sorts of complicated hacks to check if it can. And if `django-autotask` doesn't receive a resource it just makes the request normally.

So, please TopLeft and `django-autotask` to pass and accept the resource instance.

djautotask/models.py
```diff
@@ -270,7 +271,63 @@ def __str__(self):
...
+                TimeEntryTracker,
+                TicketNoteTracker,
+                TaskNoteTracker,
+            ]
+        elif self.id == \
+                self.impersonation_limited_type['CO_HELP_DESK']:
+            self.limited_impersonation_targets = [
+                Ticket,
+                Project,
+                TaskNoteTracker,
```
kti-sam, line 320: This is much better, see how you are able to use classes themselves rather than strings containing method names? This is much cleaner by comparison, and much easier to maintain than having to update a bunch of strings if a method name changes.

djautotask/models.py
```diff
@@ -164,7 +164,8 @@ def __str__(self):
 
     def update_at(self, **kwargs):
         api_client = api.TicketsAPIClient(
-            impersonation_id=kwargs.get('impersonation_id')
+            impersonation_id=kwargs.get('impersonation_id', None),
```
kti-sam, line 167: The default return value of get if a key doesn't exist is None, so you can remove all the redundant None statements added in this PR.

djautotask/api.py
kti-daniel: Done now~. Sorry for this extra review and fix. :sweat:

djautotask/models.py
kti-daniel: Yes, I agree. & I am also pleased and happy to see these. Thanks for your valuable review.

djautotask/models.py
kti-daniel: Oh,.. It is due to my recent work mixed w/ poor memory: `self.cleaned_data.pop('force_auth_check', None)`

Thanks for catching this. 

All fixed and ready for review. Hopefully no mistake from my side again.

djautotask/api.py
```diff
@@ -353,8 +355,7 @@ def get_headers(self, method):
             headers['Secret'] = self.password
         if self.integration_code:
             headers['ApiIntegrationCode'] = self.integration_code
-        if self.impersonation and self.impersonation_id \
-                and method.lower() != 'get':
+        if self.impersonation_id and method.lower() != 'get':
```
kti-sam, line 358: Seems like an extra step to convert to lowercase to compare with `get`, pretty standard to refer to request methods in captitals, like GET, POST, PATCH, etc.

djautotask/api.py
```diff
@@ -251,7 +251,8 @@ def __init__(
         integration_code=None,
         rest_api_version=None,
         server_url=None,
-        impersonation_id=None,
+        impersonation_resource=None,
+        impersonation_target=None,
```
kti-sam, line 255: Instead of using Model classes in `init_limited_impersonation_targets`, you should use API client classes. Then you could remove this `impersonation_target` argument. The client would be able to use itself, rather than relying on being passed the Model class that it doesn't actually do anything with. The API classes are 1-to-1 with models, so this should be a very easy change.

kti-daniel: I looked into that, but found some not-working cases.

We call API's w/ two different patterns for now. 

1. From Synchronizer such as `self.sctr_synchronizer(impersonation_resource=r)`, which has api client_class and it is good to go.

2. But, we have other use cases by 'save' or 'update_at' from model, which doesn't have api client_class.
e.g. 
```
card.save(impersonation_resource=r, update_at=True,
                      changed_fields=form.changed_data)
```
```
        return card.update_at(impersonation_resource=r,
                              changed_fields=updated_fields)
```

So, Using model could be consistent method to approach.

kti-daniel: Never mind... It seems possible.. Will do that.

kti-daniel: I appreciate your review again. Good to see the improvement.

PR #170, opened 2022-02-16
MOD: Convert  to

kti-daniel:


kti-sam:
The migration isn't correct. It's deleting and creating a new model. You will likely need to manually use `migrations.RenameModel`, see the docs here: https://docs.djangoproject.com/en/4.0/ref/migration-operations/#renamemodel

kti-daniel:
Billing codes, and the other records using billing codes will be synced, so isn't that okay?

My first trial was rename model, but some conflicts happened during migrations, so changed the migration method like this.

Anyway I will try that more later today or tomorrow.

PR #181, opened 2022-08-13
Update ticket synchronizer and API

kti-sam:


djautotask/sync.py
```diff
@@ -879,6 +882,43 @@ def sync_related(self, instance):
...
+    def create(self, **kwargs):
+        """
+        Make a request to Autotask to create a Ticket.
+        """
+
+        if self.client.impersonation_resource:
+            description = kwargs['description']
+        else:
+            description = "{}\n\nTicket was created by {} {}.".format(
+                kwargs['description'],
```
kti-cameron, line 911: resource and description here don't refer to fields on the ticket right? They're here purely for tracking?

kti-sam: resource is for tracking, but description is the field on the ticket.

kti-cameron: ok, my concern was what if you want to create a ticket without a resource or description. This seems to make it like a description is required to create a ticket. But maybe I'm misunderstanding it.

kti-sam: A description isn't required, but if they are creating it and not impersonating we force adding it. Originally this wasn't going to be the "form commit false" pattern, and this is leftover from that. I need to update it so it gets the description from the record generated by the form (which I intended to always have a blank description).

Updated now, I have not tested it.

kti-cameron: I see. Thanks for clarifying. I'll try the change.

PR #191, opened 2023-03-01
add owner to account model and syncs

heela-topleft:


djautotask/sync.py
```diff
@@ -1694,6 +1694,7 @@ def _assign_field_data(self, instance, object_data):
         instance.active = object_data.get('isActive')
         instance.last_activity_date = object_data.get('lastActivityDate')
         instance.phone = object_data.get('phone')
+        instance.owner_resource = object_data.get('owner_resource')
```
kti-cameron, line 1697: `owner_resource` needs be replace with `ownerResourceID`.

kti-cameron: Actually since it's a FK relationship we need to put it in `related_meta`. I'll make the change.

PR #206, opened 2023-05-14
Update README.md

zoharasulin:
I add a local development step for install django-autotask

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,16 @@ From source:
     git clone git@github.com:KerkhoffTechnologies/django-autotask.git
     cd django-autotask
     python setup.py install
+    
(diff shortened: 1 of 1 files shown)


kti-sam:
This repo doesn't have a Dockerfile. Did you mean to add one?

zoharasulin:
yes. I Didn't see that you don't have Dockerfile. can you add this?

kti-matt:
This project is a library that can be used by a Python application to integrate with Datto Autotask PSA. It doesn't run as its own application so adding a Dockerfile isn't useful.

It would be typical to use Docker to run a Python application in which this library had been installed, but that would be a problem for that app, not django-autotask.

PR #209, opened 2023-07-05
changes

heela-topleft:


djautotask/api.py
```diff
@@ -299,8 +299,10 @@ def __init__(
             rest_api_version = settings.AUTOTASK_CREDENTIALS[
                 'rest_api_version']
         if not server_url:
-            server_url = get_api_connection_url()
-
+            try:
+                server_url = get_api_connection_url()
+            except AutotaskAPIError as e:
+                raise AutotaskAPIError('{}'.format(e))
```
kti-sam, line 305: Why is this catching and reraising the same exception?

heela-topleft: fixed

djautotask/api.py
```diff
@@ -498,10 +497,14 @@ def _fetch_resource(endpoint_url, request_retry_counter=None,
...
-                                    'so this request will be retried.')
-                        raise AutotaskAPIError(response.content)
+                    try:
+                        if cached_url != \
+                                get_api_connection_url(force_fetch=True):
+                            logger.info('Zone information has been changed, '
+                                        'so this request will be retried.')
+                            raise AutotaskAPIError(response.content)
+                    except AutotaskAPIError as e:
+                        raise AutotaskAPIError('{}'.format(e))
```
kti-sam, line 507: This area originally raised an `AutotaskAPIError` and anything calling it should expect to have to handle that error, so when `get_api_connection_url` is used here there isn't a problem. If the condition is true it raises the exception, or if there is a problem in the method call it raises the same exception. So no changes need to be made here, and again you are just re-raising the caught exception so any unhandled exceptions would still unhandled.

The problem is on line 301:

```
      if not server_url:
            server_url = get_api_connection_url()
```
Where a developer might not expect that error when instantiating the class. So you need to add handling to the code that is instantiating the api class, to catch and handle the `AutotaskAPIError`. This would be easy, but it appears sentry has lost the stack trace, so we might have to track it down. If I had to guess right now (I'll look to confirm later), its in the sync code where it instantiates the API class to perform the syncs.

kti-sam:
Looks like this was actually fixed in an earlier issue

PR #216, opened 2024-03-10
Detect permissions errors in Autotask's HTTP 500 responses

kti-matt:


kti-sam:
I forgot about this. I remember when I realized that they use 500 errors for invalid credentials, I was so mad.

kti-matt:
Oh and I cleaned up the error message parsing function so it doesn't always add "No message." to the beginning of what is definitely an error message.

